Issue #4, opened 2022-08-29
Document how to retrieve results


Marr11317:
## Add a working example in readme

The current example does the test, but doesn't log nor output anything to a file. Trying it was very deceiving, although after some research, the lib seems great!

# Original issue

## Reproduction: 

```js
// test.js
import { Bench } from "tinybench";

const bench = new Bench({ time: 100 });

bench
  .add("Test tinybench", () => {
    let a = 8;
    let b = 1;
    const c = a;
    a = b;
    b = c;
  });

console.log(await bench.run());
```

```bash
node test,js
```

## Output:
```
[ m {} ]
```

## Expected output: 

Benchmark results?

Marr11317:
I've tried as CJS too, which wasn't supported because of top-level await.

```bash
$ node --version
v16.14.2
```

Marr11317:
First time using a benchmarking lib here. 

So apparently i need to listen to "cycle" event.

Are there any docs for it?

I suggest the Readme sample to be a working example.

Aslemammad:
You do not need the cycle event, it's for listening!

to see the result as mentioned in the docs, you can try accessing `Task.result` property after `bench.run`. 

You can write your understandings in the docs and I'll merge it.

Marr11317:
> You can write your understandings in the docs and I'll merge it.

sounds good

PR #5, opened 2022-08-29
doc(readme): add explanation for retrieving results

Marr11317:
see #4

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,18 +27,37 @@ You can start benchmarking by instantiating the `Bench` class and adding
 benchmark tasks to it.
 
 ```ts
-const { Bench } = require("tinybench");
+import { Bench } from 'tinybench';
+
 const bench = new Bench({ time: 100 });
 
 bench
-  .add("foo", () => {
-    // code
+  .add('switch 1', () => {
+    let a = 1;
+    let b = 2;
+    const c = a;
+    a = b;
+    b = c;
   })
-  .add("bar", async () => {
-    // code
+  .add('switch 2', () => {
+    let a = 1;
+    let b = 10;
+    a = b + a;
+    b = a - b;
+    a = b - a;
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -48,6 +50,9 @@ use it to create an another task for that instance.
 Note that the task name should always be unique in an instance, because Tinybench stores the tasks based
 on their names in a `Map`.
 
+Also note that `tinybench` does not log any result by default. You can extract the relevant stats
+from `bench.results` after running the benchmark, and process them however you want.
```
Aslemammad, line 71: `Task.results` and not bench.results.

README.md
```diff
@@ -39,6 +39,8 @@ bench
   });
 
 await bench.run();
+
+// view, print, process bench.results
```
Aslemammad: let's remove this comment here, the note is enough.

ecstrema: I believe the best would be to add an actual example of how to treat the results. 

That way the example would be complete.

Aslemammad: That's also good!

README.md
ecstrema: Well Task.result, but Bench.result**s** if I read well the documentation below.

README.md
Aslemammad: try getting the task (with bench.getTask) and log its mean or hz! that example here is good

README.md
Aslemammad: Oh i thought you just wanted to process one task result! it's okay, you can go with bench.results if you want to process all results.

README.md
Aslemammad: btw, it'd be good if we can convert the require call in the example to an import call! because it's incorrect! Thanks.

ecstrema: done

PR #9, opened 2022-08-31
ci(examples): run examples on ci to ensure they don't crash

Marr11317:
This will ensure all examples always work


Aslemammad:
Thank you so much for the hard work!

PR #13, opened 2022-09-07
fix: update package urls

nickmccurdy:
I noticed that the npm package still links to tinyspy. This should be released as a patch since it affects the docs on npm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,13 +25,13 @@
   ],
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/tinylibs/tinyspy.git"
+    "url": "git+https://github.com/tinylibs/tinybench.git"
   },
   "license": "MIT",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -23,15 +23,8 @@
...
     "dist/**"
   ],
-  "repository": {
-    "type": "git",
-    "url": "git+https://github.com/tinylibs/tinyspy.git"
-  },
+  "repository": "tinylibs/tinybench",
   "license": "MIT",
-  "bugs": {
-    "url": "https://github.com/tinylibs/tinyspy/issues"
```
nickserv, line 32: This isn't inferred from `repository`, but I don't think we really need it since there's already a link to the repository.

Aslemammad:
Thank you for the PR!

PR #14, opened 2022-09-07
chore: remove unnecessary dependencies

nickmccurdy:
When working on #13 I noticed there were some development packages copied over from tinyspy that we don't seem to use.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,6 @@
   "scripts": {
     "dev": "tsup --watch",
     "build": "tsup",
-    "prepare": "husky install",
     "publish": "npm run build && clean-publish",
     "release": "bumpp package.json --commit --push --tag && npm run publish",
     "test": "vitest --no-threads"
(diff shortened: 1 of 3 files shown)


Aslemammad:
Let's close this, since we actually need some of these dependencies! Anyways, thank you for the PR.

nickmccurdy:
Which do you need? I don't think all of them are necessary, since some of the tools like Husky aren't configured for this repo yet.

Aslemammad:
Could you config husky with eslint for this repo?

nickmccurdy:
Personally, I don't think Husky is necessary any more now that `core.hooksPath` is built into git. I've created an example of a similar hook without Husky at #15, let me know what you think.

Also, how about removing some other dependencies besides Husky?

package.json
```diff
@@ -44,15 +41,8 @@
     "eslint": "^8.22.0",
     "eslint-config-airbnb-base": "^15.0.0",
     "eslint-plugin-import": "^2.26.0",
-    "happy-dom": "^2.25.1",
-    "husky": "^7.0.4",
-    "nano-staged": "^0.5.0",
-    "prettier": "^2.5.1",
-    "size-limit": "^8.0.1",
-    "tinyspy": "*",
```
Aslemammad, line 52: feel free to only remove `tinyspy`, `prettier` and `happy-dom`!

Aslemammad:
The hook PR looks good for me if it works.

PR #16, opened 2022-09-08
chore: remove unnecessary dependencies

nickmccurdy:
Redo of #14 based on merges and feedback


Aslemammad:
Thank you so much for the hard work.

Issue #27, opened 2022-10-14
Huge difference test result compare with benchmark.js


kainstar:
When I compare the benchmark result between different libraries, I found a huge difference in 'ops/sec' value, there are my code and console result:

### code:
```js
// import Benchmark, { Suite, map, filter } from "benchmark";
import { createRequire } from 'node:module'

const require = createRequire(import.meta.url)
const Benchmark = require('benchmark')
var suite = new Benchmark.Suite();

suite
  .add("RegExp#test", function () {
    /o/.test("Hello World!");
  })
  .add("String#indexOf", function () {
    "Hello World!".indexOf("o") > -1;
  })
  // add listeners
  .on("cycle", function (event) {
    console.log(String(event.target));
  })
  .on("complete", function (benches) {
    console.log(
      "Fastest is " +
      Benchmark.map(Benchmark.filter(benches, "fastest"), "name")
    );
  })
  // run async
  .run({ async: true });
```

### result:
<img width="442" alt="image" src="https://user-images.githubusercontent.com/15249633/195813995-8d534a3e-4847-499b-adb4-5c0d24f47aac.png">

### code:
```js
import { Bench } from "tinybench";

const bench = new Bench();

bench
  .add("RegExp#test", () => {
    /o/.test("Hello World!");
  })
  .add("String#indexOf", () => {
    "Hello World!".indexOf("o") > -1;
  });

await bench.run();

console.table(
  bench.tasks.map(({ name, result }) => ({
    "Task Name": name,
    "Period (ms)": result.period,
    "Samples": result.samples.length,
    "ops/sec": result.hz,
    "Average Time (ps)": result?.mean * 1000,
    "Variance (ps)": result?.variance * 1000,
  }))
);
```

### result:
<img width="933" alt="image" src="https://user-images.githubusercontent.com/15249633/195814202-6418d824-2b0e-469b-ac1d-8616230722df.png">

You can see the benchmark.js has tens of millions ops/sec but the tinybench only has thousand.

Maybe tinybench has some extra time cost in executing bench function? Or error time statistics logic? Or there are different explanations for 'ops/sec'?

kainstar:
I make some improvement in PR #28 , maybe it helps

kainstar:
I believe I found the reason about Issue #27.

[...]

Issue #26, opened 2022-10-14
Incorrent timig with for loops


ShivamJoker:
This is my code which I am testing

```js
import { Bench } from "tinybench";

const bench = new Bench({ iterations: 100_000 });

bench
  .add("loop 1", () => {
    let sum = 1;
    for (let index = 0; indexindex < 10000; indexindex++) {
      sum = sum * index;
    }
  })
  .add("loop 2", () => {
    let sum = 1;
    for (let index = 0; indexindex < 10; indexindex++) {
      sum = sum * index;
    }
  });

await bench.run();

console.table(
  bench.tasks.map(({ name, result }) => ({
    "Task Name": name,
    "Time (ps)": result?.totalTime,
    "Variance (ps)": result?.variance * 1000,
  }))
);
```
The results are same in both

```
┌─────────┬───────────┬──────────────────────┬──────────────────────┐
│ (index) │ Task Name │      Time (ps)       │    Variance (ps)     │
├─────────┼───────────┼──────────────────────┼──────────────────────┤
│    0    │ 'loop 1'  │ 0.002133705863210011 │ 0.010551717975465412 │
│    1    │ 'loop 2'  │ 0.002180478539965521 │ 0.030143625437354347 │
└─────────┴───────────┴──────────────────────┴──────────────────────┘

```

I am using node v16.17.1

Aslemammad:
Let me know which tinybench version you are using.

And see if this is related #27 #28

Aslemammad:
ok! could you check if #28 solves your issue?

kainstar:
It's your code's error, not tinybench's problem.

Please check your for loop end condition

```
for (let index = 0; indexindex < 10000; indexindex++)
```

You declare a veriable named `index`, but the end condition is using a undeclared varibale `indexindex`.

These two code has same time because when it execute, it would throw error immediately and never into loop.

When I test correct code on v2.3.0, it give such result:
<img width="504" alt="image" src="https://user-images.githubusercontent.com/15249633/196100663-8faeaa69-9004-448b-a494-d78fdfd02f19.png">

ShivamJoker:
Aah sorry my mistake

kainstar:
Maybe tinybench think throw error is also a way to complete a function?

Aslemammad:
Please create a new issue in case! I don't think this is a problem anymore. Thanks.

PR #28, opened 2022-10-14
fix: improve task time calculate logic

kainstar:
Solve Issue #27 and do same task time impovement

## fix:
- Separate async function and normal function, avoid normal function affected by Promise micro task.
- Use the sum of samples rather than `totalTime`, because `totalTime` include some extra time in event loop.
- correct `hz`'s calculate logic

## chore:
- add eslint config for example folder

### Before
image

### After
image


package.json
```diff
@@ -3,6 +3,9 @@
   "version": "2.3.0",
   "type": "module",
   "packageManager": "pnpm@7.5.1",
+  "engines": {
+    "node": ">= 10.0.0"
```
Aslemammad: Let's remove this, since it's not only a node lib.

src/utils.ts
```diff
@@ -19,3 +19,8 @@ export const getVariance = (samples: number[], mean: number) => {
   const result = samples.reduce((sum, n) => sum + ((n - mean) ** 2));
   return result / (samples.length - 1) || 0;
 };
+
+/**
+ * Computes the sum of a sample
+ */
+export const getSum = (samples: number[]) => samples.reduce((sum, n) => sum + n, 0);
```
Aslemammad: Isn't totalTime so much faster than this?

Aslemammad: Isn't totalTime so much faster than this?

package.json
kainstar: Ok, but if it's not only a node lib, I will replace the `isAsyncFunction` with a dirty check implement because it maybe run in low version browser which is not support `async function` syntax.

src/utils.ts
kainstar: But the `getSum` will get a exact cost time (not include math calculate, array push or condition check operations).
Maybe you want sum the `totalTime` in every loop's end instead?

package.json
Aslemammad: Let's not think about those old versions, I think what mitata does is good. But anyway, if you think your solution is better, feel free to.

Aslemammad:
And please also consider this issue, as I think it's related #26

package.json
kainstar: I add a option named `dirtyAsyncCheck` in Bench's constructor, it give user a choice whether use dirty execute chcek. By default, I compare function's prototype like mitata does.

src/utils.ts
kainstar: Now I sum the `totalTime` in every loop's end

kainstar:
> And please also consider this issue, as I think it's related #26

It's not tinybench's problem, I have replyed it.

src/bench.ts
```diff
@@ -32,6 +32,8 @@ export default class Bench extends EventTarget {
 
   now = now;
 
+  isAsyncFunction: (fn: Fn) => boolean;
```
Aslemammad: This PR looks great to me.

But let's remove this option and the dirty check also! Let's only have what mitata does, because tinybench is more likely to be used in a modern environment.

kainstar: Maybe we can only remove the dirty check and keep this option so user can provide other `isAsyncFunction` implement like `now` option?

Aslemammad: Actually, It'd be good, but from my perspective, it's not reasonable and not that deeply needed! We'll do it if some folks will it, but not now. Let's remove it also, BTW, Thank you so much for your great work.

kainstar: It has removed

Aslemammad (approved):

PR #29, opened 2022-10-17
fix: rewrite sort function

### LINKED ISSUE #25: fix the sort function
the sort function in https://github.com/tinylibs/tinybench/blob/main/src/task.ts#L76 is not correct, because it only works for strings and as you may know samples are numbers! So fix it to sort numbers.
it should be something like `sort((a,b) => a - b)` I guess.

Morteza-Mazrae:
Resolves #25 
previously sample.sort() was used to sort which is wrong for numbers

### DIFF
--- a/src/task.ts
+++ b/src/task.ts
@@ -73,7 +73,7 @@ export default class Task extends EventTarget {
     }
     await this.bench.teardown(this, 'run');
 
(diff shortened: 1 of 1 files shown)


Aslemammad (approved):
LGTM!

PR #37, opened 2023-04-12
hotfix: await the this.bench.setup call during warmup

### LINKED ISSUE #36: setup hook not awaited during warmup
Currently, the `setup` method is awaited during normal benchmark runs, however it is not awaited during `warmup`, even though the setup method is per it's signature makred as potentially asynchronous.
Awaiting `setup` during warmup will create consistent behaviour for both the warmup and the actual benchmarking stage.

HlustikP:
Adds `await` for the setup method during `warmup`.

Closes #36

### DIFF
--- a/src/task.ts
+++ b/src/task.ts
@@ -172,7 +172,7 @@ export default class Task extends EventTarget {
     const startTime = this.bench.now();
     let totalTime = 0;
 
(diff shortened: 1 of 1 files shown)


Aslemammad (approved):

Aslemammad:
Thank you so much

Issue #35, opened 2023-04-08
[feat] `bench.todo()`


loynoir:
### Before
```js

if(!foobar_has_test_bug) bench.add(foobar.name, foobar)
```

- Table of `hello`, `world`, `foobar`
- Or table of `hello`, `world`

Which might let dev forget to watch upstream bug.

### After
```js
if(foobar_has_test_bug) {
  bench.todo(foobar.name, foobar)
} else {
  bench.add(foobar.name, foobar)
}
```

- Table of `hello`, `world`, `foobar`, while `foobar` maybe `N/A`

### Related

https://github.com/isomorphic-git/isomorphic-git/issues/1163#issuecomment-1500794012

Aslemammad:
Nice additions, are you interested in adding them yourself?

kamiloox:
Hey, I can try to create a PR for that. But one question - how the API could look like? Because if we want to be consistent between different versions it might not be the best thing to set a result with values of `N/A`. Marking them as `number | 'N/A'` will be a major change. My idea is to make a `todo = false` flag in Task. Example: 

![image](https://user-images.githubusercontent.com/45523480/233852794-1367d907-efd4-4e76-85ce-993734843d2d.png)
![image](https://user-images.githubusercontent.com/45523480/233852804-8d95d670-e5ff-4408-91ef-50c04a476cfd.png)

WDYT guys about this API? It's not the best but it's one of the possible solution 😄

Aslemammad:
@kamiloox I don't think any calculation is going to be there, and we just make it todo and the rest is handled by the user, no need to worry about that.

kamiloox:
@Aslemammad In the README there is an example of how to use this library. The results are multiplied by 1000, that's why I asked this question. But maybe I miss and don't see something 🤔

![image](https://user-images.githubusercontent.com/45523480/233860279-e94dc76c-4879-4d9b-9505-861700c0f759.png)

Aslemammad:
As I say, It's ok to leave it to the users. And if we do not do anything with the todo, that way the user won't even see it, so this is even better. this way it won't even be in bench.tasks

PR #39, opened 2023-04-24
feat: add bench.todo

kamiloox:
issue: https://github.com/tinylibs/tinybench/issues/35


src/bench.ts
```diff
@@ -148,6 +161,10 @@ export default class Bench extends EventTarget {
     return [...this._tasks.values()];
   }
 
+  get todos(): Task[] {
+    return [...this._todos.values()];
+  }
+
```
kamiloox, line 185: Should I create and use a new `Todo` class instead of `Task`?

Issue #41, opened 2023-06-06
Should the `before`/`after` hooks run during Task warmup?


scotttrinh:
I am benchmarking TypeScript's language service binary and I'm currently spawning the `tsserver` process in a `beforeAll` hook and then killing the process in an `afterAll` hook. However, I noticed that running the warmup does not run the task's `beforeAll` and `afterAll` hooks, but it _does_ run the bench's `setup` and `teardown` hooks. Since I'm shooting out different versions of the TypeScript compiler against each other, I cannot move my setup and teardown into the bench's setup/teardown lifecycle since each `Task` needs to setup a different server.

Here is a portion of my benchmarking code:

```ts
for (const version of ["ts48", "ts49", "ts50", "ts51"]) {
  bench = bench.add(
    version,
    async () => {
      await getCompletion(activeServer);
    },
    {
      beforeAll: () => {
        activeServer = createServer(
          path.join(
            __dirname,
            "..",
            "..",
            "..",
            "node_modules",
            version,
            "bin",
            "tsserver"
          )
        );
        activeServer.send({
          command: "open",
          arguments: {
            file: mockFileName,
            fileContent: mockFileContent,
          },
        });
      },
      afterAll: async () => {
        await activeServer.close();
      },
    }
  );
}
```

PR #43, opened 2023-06-18
fix: accurate results and hooks fix

### LINKED ISSUE #42: Results are incorrect #2
Hey, I've been trying tinybench an comparing it with Benchmark.js, and the results are super different. We're talking billions vs millions, maybe tinybench has too much overhead and that's why it fails to measure really fast functions.

I looked at the other issues and I think it's reporting weird values with simple tests as well:

```ts
bench
  .add("10", () => {
    let count = 0;
    for (let i = 0; i < 10; i++) {
      if (i % 2 === 0) count++;
    }
  })
  .add("100", () => {
    let count = 0;
    for (let i = 0; i < 100; i++) {
      if (i % 2 === 0) count++;
    }
  })
  .add("1000", () => {
    let count = 0;
    for (let i = 0; i < 1000; i++) {
      if (i % 2 === 0) count++;
    }
  });
```

```
┌─────────┬───────────┬─────────────┬────────────────────┬──────────┬─────────┐
│ (index) │ Task Name │   ops/sec   │ Average Time (ns)  │  Margin  │ Samples │
├─────────┼───────────┼─────────────┼────────────────────┼──────────┼─────────┤
│    0    │   '10'    │ '8,323,637' │ 120.13977791588471 │ '±0.68%' │ 4161819 │
│    1    │   '100'   │ '5,379,855' │ 185.87859194452994 │ '±0.36%' │ 2689928 │
│    2    │  '1000'   │ '1,378,380' │ 725.4889985173941  │ '±0.22%' │ 1000000 │
└─────────┴───────────┴─────────────┴────────────────────┴──────────┴─────────┘```

^ These are the results both with custom { time } and with { iterations: 1000000 }.
Am I using it wrong / misreading anything?
Thanks!

Aslemammad:
Resolves #42 and #41

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,6 +352,15 @@ export type BenchEvent = Event & {
 };
 ```
 
+### `process.hrtime`
+if you want more accurate results for nodejs with `process.hrtime`, then import
+	the `hrtimeNow` function from the library and pass it to the `Bench` options.
+
+```ts
+import { hrtimeNow } from 'tinybench';
+```
+It may make your benchmarks slower, check #42.
+
 ## Prior art
 
 - [Benchmark.js](https://github.com/bestiejs/benchmark.js)
--- a/src/index.ts
+++ b/src/index.ts
@@ -11,7 +11,7 @@ export type {
   BenchEvent,
 } from '../types';
 
-export { now } from './utils';
+export { now, hrtimeNow } from './utils';
 
 export { Bench, Task };
 export default Bench;
--- a/src/task.ts
+++ b/src/task.ts
@@ -1,5 +1,9 @@
 import type {
-  Fn, TaskEvents, TaskResult, TaskEventsMap, FnOptions,
+  Fn,
+  TaskEvents,
+  TaskResult,
+  TaskEventsMap,
+  FnOptions,
 } from '../types/index';
 import Bench from './bench';
 import tTable from './constants';
@@ -70,6 +74,7 @@ export default class Task extends EventTarget {
       }
 
       let taskStart = 0;
+      let taskTime = 0;
 
       try {
         taskStart = this.bench.now();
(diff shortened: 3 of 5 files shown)


Aslemammad:
@3rd could you check this one if it produces more accurate results now?

Aslemammad:
These are the results for me, it improved, but it's still not as accurate as mitata is.

image
There's always the 60-70ns overhead.
Tried removing try catch context and if elses, but it didn't work, the only thing worked is fallbacking to performance.now and making hrtime optional.

@evanwashere Any guidance on this issue?

3rd:
@Aslemammad great change, it's getting way closer now!

tseep should indeed do about 238,095,238 ops / second, but measuring that is less important than the difference between the function times, and it did improve there, the remaining error is about 10x (looking at tseep vs mitt)

This is what I get:
image

Aslemammad:
@3rd Thank you so much for your helpful insights! I tried multiple solutions but didn't work, let's stick with this one and see if anyone can make it even more faster later.

Aslemammad:
Honestly, this made my day. Thank you so much @3rd! Great job

I see, yea, I expected that the async/await is creating so much overhead since it's expected, but couldn't remove them! 

If you finalized your experiments (which work completely ok now), I'd like to see you contribute them to tinybench with not so much architecture change and any breaking change, since they'd be really helpful. The current architecture is being used in some production apps and breaking it would not look good. But I'm sure you'll find a way to make that without those issues. 
I'm completely open to getting those changes in tinybench.

Great job @3rd.

3rd:
@Aslemammad thanks, you too!
Yep, I can make a PR to port the async/sync split, but for the batching I think the benchmark time approach can't work, only the the batch size needs to be configured, maybe it could be defaulted to something like iterations / 100, but the users could still have to adjust it manually for long running tasks, and maybe it could warn

Aslemammad:
@3rd I'm excited to see the PR and the app!

PR #51, opened 2023-07-23
feat: `throws`

### LINKED ISSUE #47: Is there a reason that bench.run() doesn't throw?
I was scratching my head for a good three hours until I did this:

```js
bench.addEventListener("error", e => reportError(e))
```

Scenario:

```js
const bench = new Bench()
bench.add("i look fast but i just fail", () => { iDontExist() })
await bench.run()
console.table(bench.table())
```

Is there a reason that right now throwing inside benchmarks doesn't cause a catastrophic failure? There must  be... right? 😅

I think that it would be a good idea to make it so that either:
1. bench.run() throws
2. console.error() on any errors

Some kind of user feedback that the superfast speeds they're seeing are actually _an error_ is a good idea. 😊

Aslemammad:
Resolves #47


sirenkovladd:
Could you fix the error in [./src/event.ts](https://github.com/tinylibs/tinybench/blob/8d7bbfd0e62445215d294c190be8899d96756db5/src/event.ts#L1) ?

sirenkovladd:
Wouldn't it be a better way to use .d.ts with dist?

Aslemammad:
@sirenkovladd Hey, What do you mean?

sirenkovladd:
https://github.com/tinylibs/tinybench/blob/8d7bbfd0e62445215d294c190be8899d96756db5/src/event.ts#L1
->
```js
import type { BenchEvents } from './types';
```

Aslemammad:
@sirenkovladd thank you so much.

PR #52, opened 2023-09-04
Optimization run

sirenkovladd:
remove condition from run
check isAsync with `fn() instanceof Promise;`

### DIFF
--- a/src/task.ts
+++ b/src/task.ts
@@ -75,7 +75,11 @@ export default class Task extends EventTarget {
     await this.bench.setup(this, 'run');
 
     if (this.opts.beforeAll != null) {
-      await this.opts.beforeAll.call(this);
+      try {
+        await this.opts.beforeAll.call(this);
+      } catch (e) {
+        this.setResult({ error: e });
+      }
     }
     const isAsync = await isAsyncTask(this);
 
@@ -103,7 +107,11 @@ export default class Task extends EventTarget {
     }
 
     if (this.opts.afterAll != null) {
-      await this.opts.afterAll.call(this);
+      try {
+        await this.opts.afterAll.call(this);
+      } catch (e) {
+        this.setResult({ error: e });
+      }
     }
 
     await this.bench.teardown(this, 'run');
@@ -185,7 +193,11 @@ export default class Task extends EventTarget {
     await this.bench.setup(this, 'warmup');
 
     if (this.opts.beforeAll != null) {
-      await this.opts.beforeAll.call(this);
+      try {
+        await this.opts.beforeAll.call(this);
+      } catch (e) {
+        this.setResult({ error: e });
+      }
     }
     const isAsync = await isAsyncTask(this);
 
@@ -195,7 +207,11 @@ export default class Task extends EventTarget {
       && !this.bench.signal?.aborted
     ) {
       if (this.opts.beforeEach != null) {
-        await this.opts.beforeEach.call(this);
+        try {
+          await this.opts.beforeEach.call(this);
+        } catch (e) {
+          this.setResult({ error: e });
+        }
       }
 
       try {
@@ -213,12 +229,20 @@ export default class Task extends EventTarget {
       totalTime = this.bench.now() - startTime;
 
       if (this.opts.afterEach != null) {
-        await this.opts.afterEach.call(this);
+        try {
+          await this.opts.afterEach.call(this);
+        } catch (e) {
+          this.setResult({ error: e });
+        }
       }
     }
 
     if (this.opts.afterAll != null) {
-      await this.opts.afterAll.call(this);
+      try {
+        await this.opts.afterAll.call(this);
+      } catch (e) {
+        this.setResult({ error: e });
+      }
     }
     this.bench.teardown(this, 'warmup');
 
--- a/src/task.ts
+++ b/src/task.ts
@@ -50,20 +50,6 @@ export default class Task extends EventTarget {
     // TODO: support signals in Tasks
   }
 
-  private runSync() {
-    const taskStart = this.bench.now();
-    this.fn();
-    const taskTime = this.bench.now() - taskStart;
-    return taskTime;
-  }
-
-  private async runAsync() {
-    const taskStart = this.bench.now();
-    await this.fn();
-    const taskTime = this.bench.now() - taskStart;
-    return taskTime;
-  }
-
   /**
    * run the current task and write the results in `Task.result` object
    */
@@ -92,7 +78,16 @@ export default class Task extends EventTarget {
           await this.opts.beforeEach.call(this);
         }
 
-        const taskTime = isAsync ? await this.runAsync() : this.runSync();
+        let taskTime = 0;
+        if (isAsync) {
+          const taskStart = this.bench.now();
+          await this.fn();
+          taskTime = this.bench.now() - taskStart;
+        } else {
+          const taskStart = this.bench.now();
+          this.fn();
+          taskTime = this.bench.now() - taskStart;
+        }
 
         samples.push(taskTime);
         this.runs += 1;
--- a/src/task.ts
+++ b/src/task.ts
@@ -8,7 +8,7 @@ import type {
 import Bench from './bench';
 import tTable from './constants';
 import { createBenchEvent } from './event';
-import { getMean, getVariance, isAsyncFunction } from './utils';
+import { getMean, getVariance, isAsyncTask } from './utils';
 
 /**
  * A class that represents each benchmark task in Tinybench. It keeps track of the
@@ -77,7 +77,7 @@ export default class Task extends EventTarget {
     if (this.opts.beforeAll != null) {
       await this.opts.beforeAll.call(this);
     }
-    const isAsync = await isAsyncFunction(this);
+    const isAsync = await isAsyncTask(this);
 
     try {
       while (
@@ -187,7 +187,7 @@ export default class Task extends EventTarget {
     if (this.opts.beforeAll != null) {
       await this.opts.beforeAll.call(this);
     }
-    const isAsync = await isAsyncFunction(this);
+    const isAsync = await isAsyncTask(this);
 
     while (
       (totalTime < this.bench.warmupTime
--- a/src/task.ts
+++ b/src/task.ts
@@ -71,13 +71,13 @@ export default class Task extends EventTarget {
     this.dispatchEvent(createBenchEvent('start', this));
     let totalTime = 0; // ms
     const samples: number[] = [];
-    const isAsync = isAsyncFunction(this.fn);
 
     await this.bench.setup(this, 'run');
 
     if (this.opts.beforeAll != null) {
       await this.opts.beforeAll.call(this);
     }
+    const isAsync = await isAsyncFunction(this);
 
     try {
       while (
--- a/src/task.ts
+++ b/src/task.ts
@@ -179,7 +179,6 @@ export default class Task extends EventTarget {
    */
   async warmup() {
     this.dispatchEvent(createBenchEvent('warmup', this));
-    const isAsync = isAsyncFunction(this.fn);
     const startTime = this.bench.now();
     let totalTime = 0;
 
@@ -188,6 +187,7 @@ export default class Task extends EventTarget {
     if (this.opts.beforeAll != null) {
       await this.opts.beforeAll.call(this);
     }
+    const isAsync = await isAsyncFunction(this);
 
     while (
       (totalTime < this.bench.warmupTime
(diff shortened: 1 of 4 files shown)


src/task.ts
```diff
@@ -74,13 +88,7 @@ export default class Task extends EventTarget {
...
         }
 
-        const taskStart = this.bench.now();
-        if (isAsync) {
-          await this.fn();
-        } else {
-          this.fn();
-        }
-        const taskTime = this.bench.now() - taskStart;
+        const taskTime = isAsync ? await this.runAsync() : this.runSync();
```
kibertoad: why this change? doesn't it do exactly same thing but with more code?

src/utils.ts
```diff
@@ -25,4 +25,18 @@ const AsyncFunctionConstructor = (async () => {}).constructor;
 /**
  * an async function check method only consider runtime support async syntax
  */
-export const isAsyncFunction = (fn: Fn) => fn.constructor === AsyncFunctionConstructor;
+export const isAsyncFunction = (fn: Fn) => {
+  if (fn.constructor === AsyncFunctionConstructor) {
+    return true;
+  }
+  try {
+    const call = fn();
```
kibertoad: wouldn't that produce side effects?

sirenkovladd: Well, the function is designed to be called a large number of times, right?

src/task.ts
sirenkovladd: previously, the condition itself was included in the challenge
for quick functions it makes sense

kibertoad: What do you mean? Isn't exactly the same code going to be executed the same amount of times now, just spread differently?

sirenkovladd: In my changes, the calculation is in runAsync/runSync
I mean the execution of the condition is above the time calculation, not in the middle

kibertoad: @sirenkovladd I don't see that. Previously it calculated starting point, executed function, calculated taskTime. It does exactly same thing now.

unless you refer to `isAsync` change, which did not require the rest of the changes.

sirenkovladd: https://github.com/tinylibs/tinybench/blob/4d2c04f564586bdb3c8914fee54117a8faa0fc9e/src/task.ts#L77-L83

I see a condition in the calculation of time

src/utils.ts
sirenkovladd: Do you have one, do you have an example?

src/task.ts
sirenkovladd: If you want I can take out the `while` from the `run` in `runAsync`/`run/Sync`

Aslemammad:
I like this change, I see it removing overhead, but did you see that in the results? Have you been able to notice any thing? I'd like to see it, I'm just curious. 
This PR LGTM, but I'd like to see @mcollina and @Uzlopak's opinion on this.

src/utils.ts
```diff
@@ -25,4 +25,24 @@ const AsyncFunctionConstructor = (async () => {}).constructor;
 /**
  * an async function check method only consider runtime support async syntax
  */
-export const isAsyncFunction = (fn: Fn) => fn.constructor === AsyncFunctionConstructor;
+export const isAsyncFunction = async (task: Task) => {
+  if (task.fn.constructor === AsyncFunctionConstructor) {
```
Aslemammad: isn't this condition only enough?

Aslemammad: I guess with the new condition it determines functions that return promises? right?

sirenkovladd: Not detectable

```js
.add('foo', () => {
      return new Promise((resolve) => setTimeout(resolve, 100));
    })
```

Aslemammad: Ok, I get it now. It's reasonable.

Uzlopak:
isAsyncFunction was doing totally perfect sense, as it was fast to determine if we have a function of type `async function () {}`. 

Now isAsyncFunction makes one run of the function and determines various types of async behaviour. Maybe make it two function. Restore the previous isAsyncFunction and create a new isAsyncTask, which does all the magic.

Also, the `call instanceof Promise` is not enough. Yes it detects built-in Promises. But It should handle PromiseLike. So `call && typeof call.then === 'function'`

Uzlopak:
@Aslemammad 

Please check discord.

sirenkovladd:
> But It should handle PromiseLike So `call && typeof call.then === 'function'`

I agree with you, I will change it

src/utils.ts
```diff
@@ -25,17 +26,19 @@ const AsyncFunctionConstructor = (async () => {}).constructor;
...
   }
   try {
     if (task.opts.beforeEach != null) {
       await task.opts.beforeEach.call(task);
     }
     const call = task.fn();
-    const result = call instanceof Promise;
-    if (result) {
+    const result = typeof call?.then === 'function';
+    if (result && call?.catch) {
```
Uzlopak: You could actually send it to nirvana by doing

call.then(() => {}, () => {})

So you can save your check for catch. E.g. some fastify classes are thenable but have no .catch.

src/utils.ts
```diff
@@ -25,17 +26,19 @@ const AsyncFunctionConstructor = (async () => {}).constructor;
...
     return true;
   }
   try {
     if (task.opts.beforeEach != null) {
       await task.opts.beforeEach.call(task);
     }
     const call = task.fn();
-    const result = call instanceof Promise;
-    if (result) {
+    const result = typeof call?.then === 'function';
```
Uzlopak: Maybe `result` is too generic. Actually should be renamed to `promiseLike`.

Maybe create a typeguard isPromiseLike

See 
https://github.com/fastify/avvio/blob/master/lib/is-promise-like.js

src/utils.ts
sirenkovladd: > call.then(() => {}, () => {})

it is just `await call`

You can give some example or give some test that will reject

Uzlopak: Then why do you check for catch?

src/utils.ts
sirenkovladd: done it

src/utils.ts
sirenkovladd: Some tests throw unhandled errors

sirenkovladd:
```js
.add('error', () => {
      return Promise.reject(new Error('fake'));
    })
```

src/utils.ts
```diff
@@ -26,3 +35,29 @@ const AsyncFunctionConstructor = (async () => {}).constructor;
...
  */
 export const isAsyncFunction = (fn: Fn) => fn.constructor === AsyncFunctionConstructor;
+
+export const isAsyncTask = async (task: Task) => {
+  if (isAsyncFunction(task.fn)) {
+    return true;
+  }
+  try {
+    if (task.opts.beforeEach != null) {
+      await task.opts.beforeEach.call(task);
```
Uzlopak: hmm

kind of feels error prone. if beforeEach or afterEach fail, then we result in the fact that the Task is not async. 

imho beforeEach and afterEach have to be in try catch blocks

test/index.test.ts
```diff
@@ -316,3 +316,14 @@ test('task beforeAll, afterAll, beforeEach, afterEach', async () => {
   expect(afterEach.mock.calls.length).toBe(iterations * 2);
   expect(beforeEach.mock.calls.length).toBe(afterEach.mock.calls.length);
 });
+
+test('task with promiseLike return', async () => {
```
Uzlopak, line 320: we are lacking a test, where a sync function returns a promise

src/utils.ts
sirenkovladd: good point, thank you

test/index.test.ts
sirenkovladd: This test actually does that

Uzlopak: Why do we then bother to cover various types if we dont test for them.

This is what I want as an additional test (not tested!):

```js
test('sync task which returns a promise', async () => {
  const bench = new Bench();

  bench.add('foo', () => {
    return new Promise(resolve => setTimeout(resolve, 100))
  });

  await bench.run();
  expect(bench.getTask('foo')!.result!.mean).toBeGreaterThan(100);
});

Comment (sirenkovladd) [reply test/index.test.ts:320]: Okay, I'll add, it's the same for me

```js
bench.add('foo', () => ({
    then: async (resolve: () => void) => setTimeout(resolve, 100),
  }));
```

Uzlopak: still, add that case please.

Uzlopak:
consider creating a new test file, where you test all aspects of isAsyncTask.

Especially write unit tests, that check if isAsyncTask returns correct values if beforeEach and/or afterEach throw an error.

sirenkovladd:
@Uzlopak, that's all?

Uzlopak (approved):
LGTM

But waiting for feedback by @kibertoad and @Aslemammad

Aslemammad (approved):

sirenkovladd:
Do you have any blocks to merge PR?

Uzlopak:
I would prefer that @kibertoad have a look. I now would just say, that he has time till sunday. If on sunday no feedback is given, we should not block it longer.

src/task.ts
kibertoad: I feel that we are talking about very different things. Can we hop to discord to discuss in realtime, perhaps?

src/task.ts
```diff
@@ -95,7 +107,11 @@ export default class Task extends EventTarget {
     }
 
     if (this.opts.afterAll != null) {
-      await this.opts.afterAll.call(this);
+      try {
+        await this.opts.afterAll.call(this);
+      } catch (e) {
```
kibertoad, line 107: is this change related to optimizations?

src/utils.ts
```diff
@@ -26,3 +35,37 @@ const AsyncFunctionConstructor = (async () => {}).constructor;
...
+
+export const isAsyncTask = async (task: Task) => {
+  if (isAsyncFunction(task.fn)) {
+    return true;
+  }
+  try {
+    if (task.opts.beforeEach != null) {
+      try {
+        await task.opts.beforeEach.call(task);
+      } catch (e) {
```
kibertoad, line 47: why do we ignore thrown errors here?

sirenkovladd: Because this function is only responsible for determining whether this function is synchronous

src/task.ts
sirenkovladd: no, but because of this there will be an error in the tests

sirenkovladd:
Do you have any other questions about PR?

kibertoad:
@sirenkovladd Everything else looks good! Can we remove the unnecessary code duplication, though? I still don't see what it adds.

sirenkovladd:
Sorry, what do you mean by `unnecessary code duplication`?

kibertoad:
@sirenkovladd 

```ts
  private runSync() {
    const taskStart = this.bench.now();
    this.fn();
    const taskTime = this.bench.now() - taskStart;
    return taskTime;
  }

  private async runAsync() {
    const taskStart = this.bench.now();
    await this.fn();
    const taskTime = this.bench.now() - taskStart;
    return taskTime;
  }
```

This is exactly same code, the only difference is whether we await it or not. This can be handled as a simple condition, without copy-pasting the rest of it.

sirenkovladd:
OK, i got it
but the point is that there are no conditions when calculating the speed of execution (if i add `isSync ? call() : await call()`)
also if I call `await` when calling synchronous code (to combine these two cases) it will impose extra time on the synchronous function (already tested) which will make the calculation not correct

so I think this is the best way to calculate the execution time correctly

kibertoad:
OK, I see your point. Not sure that's the tradeoff that I would make, but I see your reasoning, and it's sound.

kibertoad (approved):

Uzlopak:
wait

Uzlopak:
just for clarification: 

we should not have a condition inside the code, where we started the benchmark. 

So something you proposed 

 ```js
const taskStart = this.bench.now(); 
if (isAsync) { 
	 await this.fn(); 
} else { 
	 this.fn(); 
}
const taskTime = this.bench.now() - taskStart; 
```

would be imho bad. Thats why I am ok with the current solution. 

What we actually just need to do is following. 

 ```js
let taskTime = 0
if (isAsync) {
    const taskStart = this.bench.now();
    await this.fn();
    taskTime = this.bench.now() - taskStart; 
} else {
    const taskStart = this.bench.now();
    this.fn();
    taskTime = this.bench.now() - taskStart; 
}
```

then we dont need the private methods runAsync and runSync, which btw. imho are not a useful abstraction.

Uzlopak:
Will ask @Aslemammad to release ;)

PR #55, opened 2023-09-11
skip empty event

sirenkovladd:
Fixes: https://github.com/tinylibs/tinybench/issues/54


sirenkovladd:
@Uzlopak @kibertoad @Aslemammad

test/index.test.ts
```diff
@@ -366,3 +366,12 @@ test('throw error in beforeAll, afterAll, beforeEach, afterEach', async () => {
...
   expect(bench.getTask('AA test')!.result!.error).toBe(AAerror);
 });
+
+test('remove non-existent task', () => {
+  const bench = new Bench();
+  bench.addEventListener('remove', () => {
+    throw new Error('should not be called');
+  });
+
+  bench.remove('non-existent');
```
Uzlopak, line 376: check how vitest tests for non throwing.

test/index.test.ts
```diff
@@ -366,3 +366,12 @@ test('throw error in beforeAll, afterAll, beforeEach, afterEach', async () => {
   expect(bench.getTask('AE test')!.result!.error).toBe(AEerror);
   expect(bench.getTask('AA test')!.result!.error).toBe(AAerror);
 });
+
+test('remove non-existent task', () => {
```
Uzlopak: try to make the description more descriptive.

```suggestion
test('removing non-existing task should not throw', () => {
```

Uzlopak:
The question, which has to be discussed before hand is, if it makes sense to ignore the non existing task, or to throw a custom error.

test/index.test.ts
sirenkovladd: No errors, just an empty event received in the listener

sirenkovladd:
I don't have much respect for throwing errors
I suggest returning boolean instead of `this`

Uzlopak:
with returning `this` we allow chaining/ train wreck pattern.

sirenkovladd:
are you sure adding `remove` to the chain is a good idea?

test/index.test.ts
sheremet-va: There is `expect.unreachable()` for situations like this

sirenkovladd: Approval required to update version for "vitest"

Uzlopak (approved):
LGTM

test/index.test.ts
Uzlopak: Update to latest vitest if necessary

PR #56, opened 2023-09-12
some improvement

sirenkovladd:
- remove useless Type Assertions
- add custom row in table
- throwing an error in table
- remove duplication code in run and warmup

### DIFF
--- a/src/bench.ts
+++ b/src/bench.ts
@@ -136,7 +136,7 @@ export default class Bench extends EventTarget {
     listener: T,
     options?: boolean | AddEventListenerOptions,
   ): void {
-    super.addEventListener(type as string, listener as any, options);
+    super.addEventListener(type, listener as any, options);
   }
 
   removeEventListener<K extends BenchEvents, T = BenchEventsMap[K]>(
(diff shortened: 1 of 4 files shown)


Uzlopak:
@sirenkovladd 

Can you resolve the conflicts, please?

sirenkovladd:
done

src/bench.ts
```diff
@@ -142,24 +142,27 @@ export default class Bench extends EventTarget {
...
   ) {
-    super.removeEventListener(type as string, listener as any, options);
+    super.removeEventListener(type, listener as any, options);
   }
 
   /**
    * table of the tasks results
    */
-  table() {
+  table(convert?: (result: TaskResult) => Record<string, string | number> | undefined) {
```
Aslemammad: Any docs?

src/task.ts
```diff
@@ -91,41 +84,51 @@ export default class Task extends EventTarget {
...
+      return { error };
     }
 
     if (this.opts.afterAll != null) {
       try {
         await this.opts.afterAll.call(this);
-      } catch (e) {
-        this.setResult({ error: e });
+      } catch (error) {
+        return { error };
```
Aslemammad, line 101: why return?

sirenkovladd: it is possible that incorrect completion of the task in afterAll will cause incorrect initialization in the next task
in which case the result cannot be trusted

That is why it was decided to return an error instead of a value

src/bench.ts
sirenkovladd: changed

Aslemammad (approved):

PR #58, opened 2023-10-04
fix: remove DOM-related types to make tinybench environment agnostic

sheremet-va:


Aslemammad (approved):

AriPerkkio (approved):
I'm not familiar with this project's codebase, but I did some testing in minimal repro with `@types/nodes` and another with `lib: ["dom"]` and both seem to work correctly with `AddEventListenerOptionsArgument`. :+1:

PR #62, opened 2023-10-23
Idea

Uzlopak:


Uzlopak:
We can load childProcess like this

```js
import createBenchEvent from "./event";
import { isESM, isNode, taskIdFromEnv } from "./utils";
import Bench from "./bench";
import Task from "./task";

let childProcess = null;

export async function runner (this: Bench) {
    this.dispatchEvent(createBenchEvent('start'));

    const taskId = taskIdFromEnv();
    if (taskId !== -1) {
        const task = this.getTask(taskId);
        if (task) {
            await task.run();
            return [task];
        }
        return [];
    }

    const values: Task[] = [];
    for (const task of [...this._tasks.values()]) {
        if (isNode) {
            console.log(isESM);
            // @ts-ignore
            childProcess ??= isESM ? await import('node:child_process') : require('node:child_process');
            const result = childProcess.execSync(`TINYBENCH_TASK_ID=${task.id} ${process.execPath}`, { env: process.env });
            console.log(result.toString())
        }

        if (this.signal?.aborted) values.push(task);
        else values.push(await task.run());
    }
    this.dispatchEvent(createBenchEvent('complete'));
    return values;
}

```

Uzlopak:
I cant get it to work properly. :D

Maybe somebody else has to take over.

src/bench.ts
```diff
@@ -175,24 +187,29 @@ export default class Bench extends EventTarget {
...
    */
-  getTask(name: string): Task | undefined {
-    return this._tasks.get(name);
+  getTask(t: string | number): Task | undefined {
+    if (typeof t === 'string') {
+      return this._tasks.get(t);
+    }
+    return [...this._tasks.values()].filter(
+      (task) => task.id === t,
+    )[0];
```
sirenkovladd: It seems like this should be a separate function, for example `getTaskById`

src/bench.ts
```diff
@@ -20,9 +20,9 @@ export default class Bench extends EventTarget {
   /*
    * @private the task map
    */
-  _tasks: Map<string, Task> = new Map();
+  _tasks: Map<string | number, Task> = new Map();
```
sirenkovladd:
```suggestion
  _tasks: Map<string, Task> = new Map();
```

src/task.ts
```diff
@@ -42,12 +42,22 @@ export default class Task extends EventTarget {
...
+  id: number;
+
   constructor(bench: Bench, name: string, fn: Fn, opts: FnOptions = {}) {
     super();
     this.bench = bench;
     this.name = name;
     this.fn = fn;
     this.opts = opts;
+    this.isTodo = opts.isTodo ?? false;
+    this.id = bench._taskIdCounter++; // eslint-disable-line no-param-reassign
```
sirenkovladd, line 59: This seems to block calling multiple benches at once if you are going to pass this id to env in the future

sirenkovladd:
Do you consider the possibility that tasks can be added in a different order at each launch?
so the `id` will be different on each run

src/bench.ts
```diff
@@ -75,6 +75,17 @@ export default class Bench extends EventTarget {
...
     const values: Task[] = [];
+
+    const taskId = taskIdFromEnv();
+    if (taskId !== -1) {
+      const task = this.getTask(taskId);
+      if (task) {
+        await task.run();
+        return [task];
+      }
+      return [];
```
sirenkovladd:
```suggestion
        await task.run();
        this.dispatchEvent(createBenchEvent('complete'));
        return [task];
      }
      this.dispatchEvent(createBenchEvent('complete'));
      return [];
```

sirenkovladd (approved):

Aslemammad:
Any summary on this PR?

jerome-benoit:
@Uzlopak @sirenkovladd: I do not see the added value of having a `todo` primitive. I will remove it.

jerome-benoit:
See #120

Issue #61, opened 2023-10-20
Excessive data retention on fast benchmarks


jdmarshall:
I don't think the warmup time for tinybench is anywhere near adequate to getting V8 to settle the code. 'Solving' that problem by deoptimizing tinybench is fighting the wrong fight.

In an attempt to get more accurate data out of tinybench, I've been using time: 2000 or in some cases 6000 for some test suites. 

The problem here is that given some leaf node tests are running 2, 3, even 4 orders of magnitude faster than the more complex tests, some of my tests are generating thousands of samples, and others are generating several million. This extra memory pressure is causing some weird race conditions with the code under test.

Now while switching the telemetry calculation to an algorithm that uses k or logn memory complexity for running totals is probably a lot to ask, retaining samples for task A while running task C is a waste of resources.

Expected behavior: Disabling all but one task should result in the same success or failure modes for the code under test. The summaries should be calculated at the end of a task and the working data discarded.

Actual behavior: all statistics are kept for the entire run.

jdmarshall:
Because the summary is calculated after teardown, I don't even have the option to force this behavior in my own tests.

Aslemammad:
I see, thank you so much for submitting an issue. 
Doesn't the `iterations` option help?

jdmarshall:
That's what I'm using as the workaround. I'm definitely still having some sort of issue with ensuring V8 is completely warm, but then that's a very old benchmarking problem.

jdmarshall:
Do you suppose it would break anyone if you calculated the performance numbers prior to teardown?

I have no way of knowing if anyone but you is depending on the existence of the telemetry data at the end of a run. I could expunge the samples during teardown. It's not awesome, but as a documented workaround I've encountered worse.

But right now teardown seems to be first, which scuttles that idea.

Aslemammad:
> Do you suppose it would break anyone if you calculated the performance numbers prior to teardown?

I don't think so, we can try and see.

Aslemammad:
> expunge the samples during teardown

The samples are the problem here?

jdmarshall:
That's where all the data goes, yes. But teardown is the last event I can see between tasks, and it is called immediately before the P## calculations are done.  I could cheese something with the setup event, but that involves figuring out what task we are on and looking up the previous task, which seems problematic.

[...]

PR #63, opened 2023-11-04
reorder teardown callback to after the error event, and setResults.

jdmarshall:
fix for #61 

Change order of operations to 
 - emit error event if it exists
 - calculate results
 - call teardown hook
 - emit cycle event

I'm not sure why you would ever call teardown before dealing with an error that was part of the test run.

Putting teardown after setResults allows a person to purge the samples to avoid memory starvation in longer test runs.

### DIFF
--- a/src/task.ts
+++ b/src/task.ts
@@ -113,9 +113,10 @@ export default class Task extends EventTarget {
       }
     }
 
-    await this.bench.teardown(this, 'run');
-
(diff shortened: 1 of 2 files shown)


jerome-benoit:
@jdmarshall: if it's still relevant, could you please update your PR?

jerome-benoit:
Fixed

PR #50, opened 2023-07-23
fix: use .call for deoptimization

Aslemammad:
Resolves #46

I tried using .call and it worked. but warmup should be used for more accurate results.

Without warmup: 
```
┌─────────┬───────────┬─────────────┬───────────────────┬──────────┬─────────┐
│ (index) │ Task Name │   ops/sec   │ Average Time (ns) │  Margin  │ Samples │
├─────────┼───────────┼─────────────┼───────────────────┼──────────┼─────────┤
│    0    │    'a'    │ '1,668,042' │ 599.505261791414  │ '±0.47%' │ 834022  │
│    1    │    'b'    │  '159,201'  │ 6281.348724802289 │ '±0.36%' │  79601  │
│    2    │    'c'    │  '162,881'  │ 6139.414994407295 │ '±0.09%' │  81441  │
└─────────┴───────────┴─────────────┴───────────────────┴──────────┴─────────┘

```

With warmup:
```
┌─────────┬───────────┬───────────┬───────────────────┬──────────┬─────────┐
│ (index) │ Task Name │  ops/sec  │ Average Time (ns) │  Margin  │ Samples │
├─────────┼───────────┼───────────┼───────────────────┼──────────┼─────────┤
│    0    │    'a'    │ '163,825' │ 6104.066796782642 │ '±0.14%' │  81913  │
│    1    │    'b'    │ '164,007' │ 6097.265856596651 │ '±0.09%' │  82004  │
│    2    │    'c'    │ '163,775' │ 6105.922276082101 │ '±0.11%' │  81888  │
└─────────┴───────────┴───────────┴───────────────────┴──────────┴─────────┘
```

PR #67, opened 2023-12-28
docs: mention `warmup` in example

kurtextrem:
see https://github.com/tinylibs/tinybench/pull/50 / comments

previously, readme didn't really mention you should use `.warmup()`.


Aslemammad:
Thank you so much

PR #85, opened 2024-08-01
fix: build correct types for CommonJS users

kanadgupta:
Hi there! My team is attempting to import this package into a CommonJS file and we're running into TypeScript compilation errors. After doing a bit of investigation, we noticed that `tinybench` uses the same types for both the CommonJS and ESM builds, which is generally discouraged.

You can check out "Are The Types Wrong?" for a handy analysis of this: https://arethetypeswrong.github.io/?p=tinybench@2.8.0

I also created a minimal reproduction of the error [in this repository](https://github.com/kanadgupta/tinybench-type-issues).

In this pull request, I made the following changes:
- [x] Upgraded `tsup` to the latest version, which will produce separate type declaration files for the CJS and ESM builds
- [x] Consolidated a few configuration options in `tsup.config.ts` and `tsconfig.json`
- [x] Updated the `exports` object in the `package.json` to correctly point to the new type declaration files. Since `tsup` colocates the `.js`. and `.d.ts` files (e.g., `index.js` has a corresponding `index.d.ts` and `index.cjs` has a corresponding `index.d.cts`), there's no need to explicitly specify the paths to the type declaration files.
- [x] Removed the `default` export from `src/index.ts` since there's already a named export. Default exports are discouraged for libraries that ship both CommonJS and ES modules and was likely part of the type issues in this repository — [see here for a great explainer on why](https://github.com/isaacs/tshy?tab=readme-ov-file#handling-default-exports).

I confirmed that these changes fixed the issues by testing this against [the "Are The Types Wrong?" CLI](https://www.npmjs.com/package/@arethetypeswrong/cli).

Before:

CleanShot 2024-08-01 at 09 28 41@2x

After:

CleanShot 2024-08-01 at 09 27 28@2x

Thanks for your work on this tool and thanks in advance for the review!

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -16,6 +16,7 @@
 		"node": true,
 		"commonjs": true
 	},
+	"ignorePatterns": ["tsup.config.ts"],
 	"rules": {
 		"import/no-extraneous-dependencies": "off",
 		"no-restricted-syntax": "off",
--- a/package.json
+++ b/package.json
@@ -14,12 +14,11 @@
   },
   "main": "./dist/index.cjs",
   "module": "./dist/index.js",
-  "types": "./dist/index.d.ts",
+  "types": "./dist/index.d.cts",
   "exports": {
-    "types": "./dist/index.d.ts",
-    "import": "./dist/index.js",
     "require": "./dist/index.cjs",
-    "default": "./dist/index.cjs"
+    "import": "./dist/index.js",
+    "default": "./dist/index.js"
   },
   "files": [
     "dist/**"
@@ -41,7 +40,7 @@
     "nano-staged": "^0.5.0",
     "p-limit": "^4.0.0",
     "size-limit": "^7.0.8",
-    "tsup": "^5.11.7",
+    "tsup": "^8.2.3",
     "typescript": "^5.2.2",
     "vite": "^2.9.12",
     "vitest": "^0.34.6"
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,10 +1,10 @@
 {
-  "strict": true,
   "compilerOptions": {
-    "target": "esnext",
+    "target": "ES2018",
     "strict": true,
     "esModuleInterop": true,
-    "moduleResolution": "node",
+    "moduleResolution": "Bundler",
+    "noEmit": true,
     "noUncheckedIndexedAccess": true,
     "baseUrl": ".",
     "skipLibCheck": true,
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,9 +1,5 @@
 lockfileVersion: '6.0'
 
-settings:
-  autoInstallPeers: true
-  excludeLinksFromLockfile: false
-
 importers:
 
   .:
@@ -51,8 +47,8 @@ importers:
         specifier: ^7.0.8
         version: 7.0.8
       tsup:
-        specifier: ^5.11.7
-        version: 5.12.9(typescript@5.2.2)
+        specifier: ^8.2.3
+        version: 8.2.3(typescript@5.2.2)
       typescript:
         specifier: ^5.2.2
         version: 5.2.2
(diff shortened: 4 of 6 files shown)


tsconfig.json
```diff
@@ -1,10 +1,10 @@
 {
-  "strict": true,
```
kanadgupta, line 2: `tsup` was flagging this because `strict` should be within the `compilerOptions` object:

CleanShot 2024-08-01 at 09 47 20@2x

tsconfig.json
```diff
@@ -1,10 +1,10 @@
...
   "compilerOptions": {
-    "target": "esnext",
+    "target": "ES2018",
     "strict": true,
     "esModuleInterop": true,
-    "moduleResolution": "node",
+    "moduleResolution": "Bundler",
+    "noEmit": true,
```
kanadgupta, line 7: Since this package is using `tsup` to trapnspile the compiled JS files:
- `noEmit` is helpful so files aren't emitted when attempting to typecheck using `tsc`
- switching the `moduleResolution` was more appropriate given that we're not transpiling the files using `tsc`

More info here: https://www.totaltypescript.com/tsconfig-cheat-sheet#not-transpiling-with-typescript

.eslintrc.json
```diff
@@ -16,6 +16,7 @@
 		"node": true,
 		"commonjs": true
 	},
+	"ignorePatterns": ["tsup.config.ts"],
```
kanadgupta, line 19: This was required in order to make ESLint pass

package.json
```diff
@@ -14,12 +14,11 @@
   },
   "main": "./dist/index.cjs",
   "module": "./dist/index.js",
-  "types": "./dist/index.d.ts",
+  "types": "./dist/index.d.cts",
   "exports": {
-    "types": "./dist/index.d.ts",
```
AriPerkkio, line 19: Doesn't this break ESM types? Where is `index.d.ts` defined now?

I would expect following to work on both cases:

```json
  "types": "./dist/index.d.ts",
  "exports": {
    ".": {
      "import": {
        "types": "./dist/index.d.ts",
        "default": "./dist/index.js"
      },
      "require": {
        "types": "./dist/index.d.cts",
        "default": "./dist/index.cjs"
      }
    },
```

kanadgupta: In this case, TypeScript will automatically infer the location of the type declaration files because they match the file names/paths of the corresopnding JS files — see these sections of their docs for more info on this:

https://www.typescriptlang.org/docs/handbook/modules/reference.html#file-extension-substitution
https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-exports

The `types` field is really only necessary if the file paths/names do not match up (e.g., the type declaration file for `./dist/index.js` is located at `./dist/some-other-file.d.ts`).

In the case of this repository and its `tsup` build process, the JS files and type declaration files have matching names/paths, meaning these two `exports` objects are semantically equivalent:

```json
"exports": {
  ".": {
    "import": {
      "types": "./dist/index.d.ts",
      "default": "./dist/index.js"
    },
    "require": {
      "types": "./dist/index.d.cts",
      "default": "./dist/index.cjs"
    }
  }
}
```

```json
"exports": {
  ".": {
    "import": "./dist/index.js",
    "require": "./dist/index.cjs"
  }
}
```

We've been [using this pattern in our `tsup`-powered libraries](https://github.com/readmeio/oas/blob/780fbadf04cbc35214c81723a5352cb7c5368e86/packages/oas/package.json#L9-L47) for a while now and have been able to load them (complete with types) into both CJS and ESM repositories with no issues.

Happy to update this PR to explicitly specify these `types` fields if you'd prefer that! In our experience, specifying the `types` has proven to be unnecessary maintenance overhead than anything else.

AriPerkkio: Nice, thanks for the links! Current change looks good to me as is.

Aslemammad (approved):
Thank you so much for the effort!

PR #172, opened 2024-10-29
fix: fix browser support

### LINKED ISSUE #171: ReferenceError in browsers with v3.0.0
Error: "ReferenceError: process is not defined

Repro: https://jsbin.com/sicotarebe/1/edit?html,console

jerome-benoit:
closes #171

### DIFF
--- a/src/task.ts
+++ b/src/task.ts
@@ -104,8 +104,7 @@ export default class Task extends EventTarget {
       const promises: Promise<void>[] = []; // only for task level concurrency
       while (
         (totalTime < time
-          || samples.length + limit.activeCount + limit.pendingCount
-            < iterations)
+          || samples.length + limit.activeCount + limit.pendingCount < iterations)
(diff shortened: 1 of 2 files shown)


dasa:
Alternative could be to feature detect using something like this:

```ts
  if ((globalThis as any).process?.hrtime?.bigint) {
    //
  }
```

Thanks for the quick fix!

jerome-benoit:
> Alternative could be to feature detect using something like this:

I plan to add the JS runtime as a `Bench` property later on. And also add it to the results object.

dasa:
With v3.0.1 the error is:

> Uncaught TypeError: Cannot read properties of undefined (reading 'versions')

jerome-benoit:
> With v3.0.1 the error is:
> 
> > Uncaught TypeError: Cannot read properties of undefined (reading 'versions')

Fixed in 3.0.3. The linter removed overzealously null condition checks.

PR #210, opened 2024-12-21
feat: add `runSync` method to `Bench` to force benchmarks to be synchronous

### LINKED ISSUE #202: Support for synchronous benchmarks
Hey folks! I'm an engineer working on React Native at Meta and we're doing some explorations about benchmarking React Native internals (this isn't meant for developers using React Native, but for the people maintaining React Native itself).

We started these explorations using `tinybench` but we have some limitations requiring parts of these tests to be fully synchronous (so no async functions or promises). Would you be open to a PR to support this use case? Basically a way to call `bench.run` so it runs them synchronously and returns `void` or the benchmark results directly, instead of a promise.

In that case, do you have any preferences on how to set that up? Some options I could think of are:
1. Add a new `runSync` method that runs synchronously and ensures none of the configuration methods return promises.
2. Make `run` return a promise or not based on whether any of the hooks provided promises. This feels like it could be complex to implement and it would easily switch from sync to async accidentally.

If you think this is something that would be valuable for your users overall, I'd happy to submit a PR implementing it. Otherwise we can set up an internal fork for this use case, so no pressure :).

Thanks!

rubennorte:
It also validates that all the hooks (beforeAll, beforeEach, etc.) are synchronous as well and makes the benchmark fail if any of them is async (returns a promise-like object).

I did a slight refactor to extract logic common to both sync and async, and duplicated the tests that made sense to test for flavors.

closes #202

### DIFF
--- a/src/bench.ts
+++ b/src/bench.ts
@@ -20,7 +20,7 @@ import {
 } from './constants'
 import { createBenchEvent } from './event'
 import { Task } from './task'
-import { type JSRuntime, mToNs, now, runtime, runtimeVersion } from './utils'
+import { invariant, type JSRuntime, mToNs, now, runtime, runtimeVersion } from './utils'
 
 /**
  * The Bench class keeps track of the benchmark tasks and controls them.
@@ -207,6 +207,20 @@ export class Bench extends EventTarget {
     return values
   }
 
+  runSync (): Task[] {
+    invariant(this.concurrency === null, 'Cannot use `concurrency` option when using `runSync`')
+    if (this.opts.warmup) {
+      this.warmupTasksSync()
+    }
+    const values: Task[] = []
+    this.dispatchEvent(createBenchEvent('start'))
+    for (const task of this._tasks.values()) {
+      values.push(task.runSync())
+    }
+    this.dispatchEvent(createBenchEvent('complete'))
+    return values
+  }
+
   /**
    * table of the tasks results
    * @param convert - an optional callback to convert the task result to a table record
@@ -256,6 +270,16 @@ export class Bench extends EventTarget {
       for (const task of this._tasks.values()) {
         await task.warmup()
       }
+    }
+  }
+
+  /**
+   * warmup the benchmark tasks (sync version)
+   */
+  private warmupTasksSync (): void {
+    this.dispatchEvent(createBenchEvent('warmup'))
+    for (const task of this._tasks.values()) {
+      task.warmupSync()
     }
   }
 }
--- a/src/task.ts
+++ b/src/task.ts
@@ -12,7 +12,7 @@ import type {
 } from './types'
 
 import { createBenchEvent, createErrorEvent } from './event'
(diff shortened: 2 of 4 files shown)


src/task.ts
```diff
@@ -278,6 +269,69 @@ export class Task extends EventTarget {
...
+    if (this.fnOpts.afterAll != null) {
+      try {
+        const afterAllResult = this.fnOpts.afterAll.call(this)
+        invariant(!isPromiseLike(afterAllResult), 'afterAll function must be sync when using runSync()')
+      } catch (error) {
+        return { error }
+      }
+    }
+    return { samples }
+  }
```
jerome-benoit, line 334: I wonder if the sync and async implementation can share more code. 
The issue is lying in that complete asyncness detection requires to run a function at least once. And the hooks are not expected to be run twice.

test/index.test.ts
```diff
@@ -3,10 +3,20 @@ import { expect, test, vi } from 'vitest'
...
 import { Bench, hrtimeNow, now, type Task } from '../src'
 
+/**
+ * @param ms amount of time to sleep in milliseconds
+ */
+function sleep (ms: number): void {
+  const start = performance.now()
+  while (performance.now() - start < ms) {
+    // noop
+  }
```
jerome-benoit, line 13: That `sleep` implementation will hog the event loop. Do you need the timing precision brought here versus a promise based `sleep`?

jerome-benoit (commented):
Looks good at a first glance.

src/bench.ts
```diff
@@ -207,6 +207,19 @@ export class Bench extends EventTarget {
     return values
   }
 
+  runSync (): Task[] {
+    if (this.opts.warmup) {
```
jerome-benoit, line 212: An error about not supporting concurrent benchmark (`concurrency` set to a non `null` value) should be issued.

src/task.ts
rubennorte: I think it's just better to have a more explicit API that sets expectations, and hooks are not executed outside the tests themselves.

test/index.test.ts
rubennorte: We can't have a promise based sleep here precisely because the tests are synchronous. We have to either spend the time in the tests or mock timers for the same reason.

src/bench.ts
```diff
@@ -207,6 +207,22 @@ export class Bench extends EventTarget {
     return values
   }
 
+  runSync (): Task[] {
+    if (this.concurrency != null) {
+      throw new Error('Cannot use `concurrency` option when using `runSync`')
+    }
```
jerome-benoit: Please do the same for the `runSync` method in the `Task` class and we're good to go (yes, it's bad to have `Task` exported not only as a type but unfortunately the API have already been widely abused)

jerome-benoit (approved):
Ready to land

rubennorte:
Thanks for the reviews, @jerome-benoit !!

Issue #231, opened 2025-01-28
Discussion: making the result table more compact


pallosp:
My workflow includes running benchmarks in the VS Code terminal regularly.

Currently I can't give the tasks longer names than 20 characters, because they would make the table overflow and break its layout. Therefore I'm suggesting a couple of changes to make the metric columns narrower.

#### (index)

- Rename the column to `#`. Saves 5-6 characters.

#### Latency average

- Limit its precision. The digits after the 5th one are just noise. While this won't save more than 1-2 characters, it will make the numbers easier to read.
- Optional: rename `average` to `avg.`

#### Latency median

- Rename `median` to `50%`.
- Drop the ± part from the numbers. I don't think they give any actionable insight.
- Drop the trailing zeros. Round to integer when the median has too many digits.
- Return numbers, not strings.

These changes together may save 3-7 characters.

#### Throughput average

- Optional: rename `average` to `avg.`

#### Throughput median

- Rename `median` to `50%`. Saves 3 characters.
- Return numbers, not strings.

#### Before
```
Inserting 50k elements into various data structures
┌─────────┬──────────────────────┬──────────────────────┬───────────────────────┬────────────────────────────┬───────────────────────────┬─────────┐
│ (index) │ Task name            │ Latency average (ns) │ Latency median (ns)   │ Throughput average (ops/s) │ Throughput median (ops/s) │ Samples │
├─────────┼──────────────────────┼──────────────────────┼───────────────────────┼────────────────────────────┼───────────────────────────┼─────────┤
│ 0       │ 'float keyed Map'    │ '3218139.39 ± 3.66%' │ '2756229.00 ± 146.00' │ '331 ± 2.19%'              │ '363'                     │ 312     │
│ 1       │ 'int keyed Map'      │ '2211478.20 ± 1.77%' │ '2076166.00'          │ '462 ± 1.06%'              │ '482'                     │ 453     │
│ 2       │ '10 int keyed Maps'  │ '1304029.77 ± 1.12%' │ '1218542.00'          │ '782 ± 0.88%'              │ '821'                     │ 767     │
│ 3       │ 'sparse Array, 0.5'  │ '147581.90 ± 1.37%'  │ '124250.00'           │ '7543 ± 0.44%'             │ '8048'                    │ 6776    │
│ 4       │ 'sparse Array, 0.1'  │ '655357.51 ± 0.44%'  │ '654250.00'           │ '1537 ± 0.43%'             │ '1528'                    │ 1526    │
[...]

PR #233, opened 2025-01-28
feat: improve the latency and throughput formatting in the result table

pallosp:
Make the result table a bit more compact, as per the discussion under issue #231

- Reduce the number of latency and throughput digits, because the least significant ones were just noise.
- Drop the ± part from the medians. They indicated the distance between the two middle elements when there was even number of them, which isn't particularly useful.
- Output the medians as numbers to get rid of the quotes.

Example result table for `simple benchmark gc`:

```
┌─────────┬───────────────┬──────────────────────┬─────────────────────┬────────────────────────────┬───────────────────────────┬─────────┐
│ (index) │ Task name     │ Latency average (ns) │ Latency median (ns) │ Throughput average (ops/s) │ Throughput median (ops/s) │ Samples │
├─────────┼───────────────┼──────────────────────┼─────────────────────┼────────────────────────────┼───────────────────────────┼─────────┤
│ 0       │ 'faster task' │ '3578.7 ± 2.83%'     │ 3125                │ '314227 ± 0.18%'           │ 320000                    │ 27944   │
│ 1       │ 'slower task' │ '1164166 ± 3.01%'    │ 1192251             │ '1321 ± 67.29%'            │ 839                       │ 86      │
└─────────┴───────────────┴──────────────────────┴─────────────────────┴────────────────────────────┴───────────────────────────┴─────────┘
```

### DIFF
--- a/src/bench.ts
+++ b/src/bench.ts
@@ -21,6 +21,7 @@ import {
 import { createBenchEvent } from './event'
 import { Task } from './task'
 import {
+  formatNumber,
   invariant,
   type JSRuntime,
   mToNs,
@@ -254,12 +255,12 @@ export class Bench extends EventTarget {
             }
           : (convert?.(task) ?? {
               'Task name': task.name,
-              'Latency average (ns)': `${mToNs(task.result.latency.mean).toFixed(2)} \xb1 ${task.result.latency.rme.toFixed(2)}%`,
+              'Latency average (ns)': `${formatNumber(mToNs(task.result.latency.mean), 5, 2)} \xb1 ${task.result.latency.rme.toFixed(2)}%`,
               // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-              'Latency median (ns)': `${mToNs(task.result.latency.p50!).toFixed(2)}${Number.parseFloat(mToNs(task.result.latency.mad!).toFixed(2)) > 0 ? ` \xb1 ${mToNs(task.result.latency.mad!).toFixed(2)}` : ''}`,
+              'Latency median (ns)': +formatNumber(mToNs(task.result.latency.p50!), 5, 2),
               'Throughput average (ops/s)': `${task.result.throughput.mean.toFixed(0)} \xb1 ${task.result.throughput.rme.toFixed(2)}%`,
               // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-              'Throughput median (ops/s)': `${task.result.throughput.p50!.toFixed(0)}${Number.parseInt(task.result.throughput.mad!.toFixed(0), 10) > 0 ? ` \xb1 ${task.result.throughput.mad!.toFixed(0)}` : ''}`,
+              'Throughput median (ops/s)': Math.round(task.result.throughput.p50!),
               Samples: task.result.latency.samples.length,
             })
         /* eslint-enable perfectionist/sort-objects */
--- a/src/bench.ts
+++ b/src/bench.ts
@@ -260,7 +260,7 @@ export class Bench extends EventTarget {
               'Latency median (ns)': `${formatNumber(mToNs(task.result.latency.p50!), 5, 2)} \xb1 ${mToNs(task.result.latency.mad!).toFixed(2)}`,
               'Throughput average (ops/s)': `${task.result.throughput.mean.toFixed(0)} \xb1 ${task.result.throughput.rme.toFixed(2)}%`,
               // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-              'Throughput median (ops/s)': Math.round(task.result.throughput.p50!),
+              'Throughput median (ops/s)': `{Math.round(task.result.throughput.p50!)} \xb1 ${Math.round(task.result.throughput.mad)}`,
               Samples: task.result.latency.samples.length,
             })
         /* eslint-enable perfectionist/sort-objects */
--- a/src/bench.ts
+++ b/src/bench.ts
@@ -257,7 +257,7 @@ export class Bench extends EventTarget {
               'Task name': task.name,
               'Latency average (ns)': `${formatNumber(mToNs(task.result.latency.mean), 5, 2)} \xb1 ${task.result.latency.rme.toFixed(2)}%`,
               // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-              'Latency median (ns)': +formatNumber(mToNs(task.result.latency.p50!), 5, 2),
+              'Latency median (ns)': `${formatNumber(mToNs(task.result.latency.p50!), 5, 2)} \xb1 ${mToNs(task.result.latency.mad!).toFixed(2)}`,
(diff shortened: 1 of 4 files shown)


src/bench.ts
```diff
@@ -254,12 +255,12 @@ export class Bench extends EventTarget {
             }
           : (convert?.(task) ?? {
               'Task name': task.name,
-              'Latency average (ns)': `${mToNs(task.result.latency.mean).toFixed(2)} \xb1 ${task.result.latency.rme.toFixed(2)}%`,
+              'Latency average (ns)': `${formatNumber(mToNs(task.result.latency.mean), 5, 2)} \xb1 ${task.result.latency.rme.toFixed(2)}%`,
               // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-              'Latency median (ns)': `${mToNs(task.result.latency.p50!).toFixed(2)}${Number.parseFloat(mToNs(task.result.latency.mad!).toFixed(2)) > 0 ? ` \xb1 ${mToNs(task.result.latency.mad!).toFixed(2)}` : ''}`,
```
jerome-benoit, line 259: Please keep the threshold condition here : when the median absolute deviation is upper zero, it's a sign of statistical significance issue during the benchmark and a valuable information. 

You can use the same helper to display it.

src/bench.ts
```diff
@@ -254,12 +255,12 @@ export class Bench extends EventTarget {
             }
           : (convert?.(task) ?? {
               'Task name': task.name,
-              'Latency average (ns)': `${mToNs(task.result.latency.mean).toFixed(2)} \xb1 ${task.result.latency.rme.toFixed(2)}%`,
+              'Latency average (ns)': `${formatNumber(mToNs(task.result.latency.mean), 5, 2)} \xb1 ${task.result.latency.rme.toFixed(2)}%`,
```
jerome-benoit, line 258: I think you can also use the formaNumber helper for the standard deviation.

Displaying number only in the column would be better, but I do not think it's possible because of string only characters usage.

jerome-benoit (changes requested):

src/bench.ts
pallosp: I'm not sure whether I get correctly what the absolute deviation is supposed to measure.

So far I've only seen it in the table when the number of samples was even. If the median for even number of samples is problematic, shouldn't the task be executed one more time instead of reporting a problem?

src/bench.ts
pallosp: For percentages less than 1000% the helper returns the same string as `toFixed(2)`. I slightly prefer keeping `toFixed(2)` because of its simpler semantics.

src/bench.ts
jerome-benoit: It's totally unrelated to the number of samples. It's a robust statistical indicator of the statistical significance of the median value. And the only one for median. Like the standard deviation (or the absolute mean deviation) is for the average.

Adding a threshold to it is only way to reduce the information to display.

pallosp: Thanks for the explanation. After some debugging I understood what's going on. There is a bug in `absoluteDeviation` that makes `mad` exactly zero as opposed to a small positive number in most cases. I filed #234.

src/bench.ts
```diff
@@ -254,12 +255,12 @@ export class Bench extends EventTarget {
...
               // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-              'Latency median (ns)': `${mToNs(task.result.latency.p50!).toFixed(2)}${Number.parseFloat(mToNs(task.result.latency.mad!).toFixed(2)) > 0 ? ` \xb1 ${mToNs(task.result.latency.mad!).toFixed(2)}` : ''}`,
+              'Latency median (ns)': +formatNumber(mToNs(task.result.latency.p50!), 5, 2),
```
jerome-benoit:
```suggestion
              'Latency median (ns)': `${formatNumber(mToNs(task.result.latency.p50!), 5, 2)} \xb1 ${mToNs(task.result.latency.mad!).toFixed(2)}`,
```

src/bench.ts
```diff
@@ -254,12 +255,12 @@ export class Bench extends EventTarget {
...
               'Throughput average (ops/s)': `${task.result.throughput.mean.toFixed(0)} \xb1 ${task.result.throughput.rme.toFixed(2)}%`,
               // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-              'Throughput median (ops/s)': `${task.result.throughput.p50!.toFixed(0)}${Number.parseInt(task.result.throughput.mad!.toFixed(0), 10) > 0 ? ` \xb1 ${task.result.throughput.mad!.toFixed(0)}` : ''}`,
+              'Throughput median (ops/s)': Math.round(task.result.throughput.p50!),
```
jerome-benoit:
```suggestion
              'Throughput median (ops/s)': `{Math.round(task.result.throughput.p50!)} \xb1 ${Math.round(task.result.throughput.mad)}`,
```

test/index.test.ts
```diff
@@ -117,11 +117,11 @@ test('bench table (async)', async () => {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
       'Latency average (ns)': expect.any(String),
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      'Latency median (ns)': expect.any(String),
+      'Latency median (ns)': expect.any(Number),
```
jerome-benoit:
```suggestion
      'Latency median (ns)': expect.any(String),
```

test/index.test.ts
```diff
@@ -117,11 +117,11 @@ test('bench table (async)', async () => {
...
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
       'Throughput average (ops/s)': expect.any(String),
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      'Throughput median (ops/s)': expect.any(String),
+      'Throughput median (ops/s)': expect.any(Number),
```
jerome-benoit:
```suggestion
      'Throughput median (ops/s)': expect.any(String),
```

test/index.test.ts
```diff
@@ -164,11 +164,11 @@ test('bench table (sync)', () => {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
       'Latency average (ns)': expect.any(String),
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      'Latency median (ns)': expect.any(String),
+      'Latency median (ns)': expect.any(Number),
```
jerome-benoit:
```suggestion
      'Latency median (ns)': expect.any(String),
```

test/index.test.ts
```diff
@@ -164,11 +164,11 @@ test('bench table (sync)', () => {
...
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
       'Throughput average (ops/s)': expect.any(String),
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      'Throughput median (ops/s)': expect.any(String),
+      'Throughput median (ops/s)': expect.any(Number),
```
jerome-benoit:
```suggestion
      'Throughput median (ops/s)': expect.any(String),
```

PR #253, opened 2025-03-07
feat!: esm-only

43081j:
Moves to being esm-only.

Unlike other tinylib projects, this one keeps `tsup` since we bundle dependencies still.

If we ever decide not to bundle dependencies, we should drop tsup and ship the typescript emit as-is.

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -2,9 +2,5 @@
   {
     "path": "dist/index.js",
     "limit": "12 kB"
-  },
-  {
-    "path": "dist/index.cjs",
-    "limit": "12 kB"
   }
 ]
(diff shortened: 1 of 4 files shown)


jerome-benoit:
Thanks, the CI action checking the size also needs CJS handling removal.

43081j:
Any idea how we can make it happy? 

It seems confused since we removed a file

jerome-benoit:
> Any idea how we can make it happy?
> 
> It seems confused since we removed a file

It's not a file removal, it's an action configuration cleanup. I can take over the PR since it a branch on the upstream GitHub repo.

43081j:
Interesting. For future reference, where does that live?

43081j:
i've updated the config 👍 still waiting on the other tinylibs repos anyway so lets coordinate releases on discord

jerome-benoit (approved):

PR #259, opened 2025-03-17
feat: add mode to fn hooks

crimx:
Provide a way to check the mode for fn hooks that is consistent with setup and teardown.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -114,32 +114,41 @@ export type EventListener = (evt: BenchEvent) => void
 export type Fn = () => Promise<unknown> | unknown
 
 /**
+ * The task hook function signature
+ * If warmup is enabled, the hook will be called twice, once for the warmup and once for the run.
+ * @param task the task instance
+ * @param mode the mode where the hook is being called
+ */
+export type FnHook = (this: Task, mode: 'run' | 'warmup') => Promise<void> | void
+
+/**
  * the task function options
  */
 export interface FnOptions {
   /**
    * An optional function that is run after all iterations of this task end
    */
-  afterAll?: (this: Task, mode: 'run' | 'warmup') => Promise<void> | void
+  afterAll?: FnHook
 
   /**
    * An optional function that is run after each iteration of this task
    */
-  afterEach?: (this: Task, mode: 'run' | 'warmup') => Promise<void> | void
+  afterEach?: FnHook
 
   /**
    * An optional function that is run before iterations of this task begin
    */
-  beforeAll?: (this: Task, mode: 'run' | 'warmup') => Promise<void> | void
+  beforeAll?: FnHook
 
   /**
    * An optional function that is run before each iteration of this task
    */
-  beforeEach?: (this: Task, mode: 'run' | 'warmup') => Promise<void> | void
+  beforeEach?: FnHook
 }
 
 /**
  * Hook function signature
+ * If warmup is enabled, the hook will be called twice, once for the warmup and once for the run.
  * @param task the task instance
  * @param mode the mode where the hook is being called
  */
--- a/src/types.ts
+++ b/src/types.ts
@@ -116,10 +116,9 @@ export type Fn = () => Promise<unknown> | unknown
 /**
  * The task hook function signature
  * If warmup is enabled, the hook will be called twice, once for the warmup and once for the run.
- * @param task the task instance
  * @param mode the mode where the hook is being called
  */
-export type FnHook = (this: Task, mode: 'run' | 'warmup') => Promise<void> | void
+export type FnHook = (this: Task, mode?: 'run' | 'warmup') => Promise<void> | void
 
 /**
  * the task function options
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -120,22 +120,22 @@ export interface FnOptions {
...
    * An optional function that is run before iterations of this task begin
    */
-  beforeAll?: (this: Task) => Promise<void> | void
+  beforeAll?: (this: Task, mode: 'run' | 'warmup') => Promise<void> | void
 
   /**
    * An optional function that is run before each iteration of this task
    */
-  beforeEach?: (this: Task) => Promise<void> | void
+  beforeEach?: (this: Task, mode: 'run' | 'warmup') => Promise<void> | void
```
jerome-benoit: Please reuse the existing Hook type in that file. 

Update: factor out a type definition, Hook covers another need.

crimx: I followed the original code to avoid naming a new type😁. What name do you suggest? `FnHook`?

crimx:
> The idea and initial implementation are ok, but lack:
> 
> * existing JSDoc update
> * documentation update
> * dedicated unit tests testing that callbacks mode is respected
> 
> Since it's a breaking change in the current PR implementation, a backward compatible implementation should be investigated.

- I noticed the JSDoc artifacts are in the source code. Should I build the JSDoc manually? That would be hard to review.
- There is not existing documentation to update.
- Tests added, just like `setup` and `teardown`.

src/types.ts
jerome-benoit: > I followed the original code to avoid naming a new type😁. What name do you suggest? `FnHook`?

Looks like a fine name. But make the `mode` argument optional to avoid breaking TS tinybench usage code base (or JS code base using type checking):

```ts
export type FnHook<Task extends Task> = (
  this?: Task,
  mode?: 'run' | 'warmup')
) => Promise<void> | void
```

crimx:
I don't see any breaking change? The callbacks are backward compatible. It is unlikely that user passed in a callback with an optional parameter.

jerome-benoit:
> * I noticed the JSDoc artifacts are in the source code. Should I build the JSDoc manually? That would be hard to review.
Since the existing code is not documented, only the public API change should have a JSDoc, so the `FnHook` type definition should have a typedoc compatible comment. 
I will probably add a proper one on `Hook` type definition. 
> * There is not existing documentation to update.
I guess a properly commented `FnHook` type definition should be enough.

jerome-benoit:
> I don't see any breaking change? The callbacks are backward compatible. It is unlikely that user passed in a callback with an optional parameter.

See https://github.com/tinylibs/tinybench/pull/259#discussion_r1999878649

crimx:
> > I don't see any breaking change? The callbacks are backward compatible. It is unlikely that user passed in a callback with an optional parameter.
> 
> See [#259 (comment)](https://github.com/tinylibs/tinybench/pull/259#discussion_r1999878649)

You don't need to make `mode` optional. Types added in callbacks are backward compatible. [Playground Link](https://www.typescriptlang.org/play/?#code/MYGwhgzhAEAqkGtoG8BQqC+6CWA7ALgKYBOAZmMIdAPYgAmA8gA74xrQfQBGhp1xhAIIgQACnwALbBABccRAEo5AN2rY6mHARLlK0XIQDuzVilSduvfkJHips+RAQAaaAFtqdQnIDkxAK64PtAAPtA+hmDEbv5MPkrQquqaqKSBwPjY1Lg09IJ0dKLULA60jCUKZilpuBlZOQaG+YXFrHKNJhCVaCllzaLsnDx8AsJi3egWWBgK6I39gxzD1mOiE+ac0wpAA)

jerome-benoit:
> > See [#259 (comment)](https://github.com/tinylibs/tinybench/pull/259#discussion_r1999878649)
> 
> You don't need to make `mode` optional. Parameters added in callbacks are backward compatible. [Playground Link](https://www.typescriptlang.org/play/?#code/MYGwhgzhAEAqkGtoG8BQqC+6CWA7ALgKYBOAZmMIdAPYgAmA8gA74xrQfQBGhp1xhAIIgQACnwALbBABccRAEo5AN2rY6mHARLlK0XIQDuzVilSduvfkJHips+RAQAaaAFtqdQnIDkxAK64PtAAPtA+hmDEbv5MPkrQquqaqKSBwPjY1Lg09IJ0dKLULA60jCUKZilpuBlZOQaG+YXFrHKNJhCVaCllzaLsnDx8AsJi3egWWBgK6I39gxzD1mOiE+ac0wpAA)

It will fail with strict type checking options enabled.

crimx:
image

No I believe it won't. The type would be broken in these two scenarios:

1. If users directly utilize the `FnHook` type for their definitions, but this type wasn't previously available, so it won't occur.
2. If users provide callbacks with optional parameters, which is unlikely given the context of how opts is used.

src/types.ts
crimx: done

jerome-benoit (approved):

jerome-benoit:
Thanks !

PR #281, opened 2025-05-05
refactor: switch to tsdown

sxzz:
https://github.com/e18e/ecosystem-issues/issues/170


Aslemammad:
i see the tsdown takeover there! let's goooo.

tsdown.config.ts
```diff
@@ -0,0 +1,9 @@
+import { defineConfig } from 'tsdown'
+
+export default defineConfig({
+  entry: ['src/index.ts'],
+  minify: {
+    compress: true,
+    mangle: true,
+  },
+})
```
jerome-benoit, line 9: I guess the tunables mirror the tsup ones.

tsdown.config.ts
```diff
@@ -0,0 +1,9 @@
+import { defineConfig } from 'tsdown'
```
jerome-benoit:
```suggestion
import { defineConfig } from 'tsdown/config'
```

jerome-benoit (approved):

PR #287, opened 2025-05-11
chore!: drop EOLed node 18.x support

jerome-benoit:


43081j:
good change jerome! does this mean a new major?

jerome-benoit:
> good change jerome! does this mean a new major?

Since I'm maintaining several FOSS projects, I have only limited time to share between them. And I'm not sure that PR contains an awaited change. The next time I can do coding work on tinybench, I think I will finish the tinybench upgrade in vitest PR that looks like more awaited.

PR #364, opened 2025-10-18
feat: add AbortSignal support at the task level

### LINKED ISSUE #139: Feature: add abort signal support at the task level

boneskull:
Fixes #139

This change enables `Task` to be aborted via an `AbortSignal` which falls back to a signal its
`Bench`'s signal.

It adds a `signal` prop to the `Task` class, as well as an `isAborted()` method (which could be a
getter, if you wish).

I'm wrapping `tinybench` and have confirmed this works well, as I'm able to abort (async) tasks via
the signal.

I've added tests covering intended usage and some edge cases, but it seems there may be a
memory-leak issue with concurrent tasks (I'm not sure if this is a known bug); this test
(`task-level abort: works with task concurrency`) is skipped.  If you can get it working, then
maybe it's just my environment.

I added documentation about the bench-and-task-level-`AbortSignal` support to `README.md`.  Happy
to remove if it's not needed--or I can remove the examples, or whatever.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,107 @@ bench.concurrency = 'task' // The concurrency mode to determine how tasks are ru
 await bench.run()
 ```
 
+## Aborting Benchmarks
+
+Tinybench supports aborting benchmarks using `AbortSignal` at both the bench and task levels:
+
+### Bench-level Abort
+
+Abort all tasks in a benchmark by passing a signal to the `Bench` constructor:
+
+```ts
+const controller = new AbortController()
+
+const bench = new Bench({ signal: controller.signal })
+
+bench
+  .add('task1', () => {
+    // This will be aborted
+  })
+  .add('task2', () => {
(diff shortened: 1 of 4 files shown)


43081j:
looks like a great change to me!

i'd really love to ditch p-limit, its almost certainly the source of a whole bunch of garbage collection and will be contributing to your memory leak findings. we should be using our own concurrency function given we're a benchmarking library IMO (one we can control and... write better).

what do you think @jerome-benoit ?  in a separate PR some time

edit: have opened #365 and it doesn't seem to hit the same OOM exception when i try it locally

jerome-benoit:
The feature implementation looks good. Please address the typing issue in tests triggering a CI failure. 

Thanks.

boneskull:
OK, fixed the type error. Thanks for looking at this so quickly!

PR #547, opened 2026-05-07
build: do not minify published code

AriPerkkio:
Published code should not be minified. It has many downsides, like making debugging more difficult, making it easier to hide malicious code there, etc. End users can minify it if they are bundling Tinybench anywhere.

### DIFF
--- a/tsdown.config.ts
+++ b/tsdown.config.ts
@@ -2,10 +2,6 @@ import { defineConfig } from 'tsdown/config'
 
 export default defineConfig({
   entry: ['src/index.ts'],
-  minify: {
-    compress: true,
-    mangle: true,
-  },
   outExtensions: () => {
     return {
       dts: '.d.ts',
--- a/tsdown.config.ts
+++ b/tsdown.config.ts
@@ -8,4 +8,7 @@ export default defineConfig({
       js: '.js',
     }
   },
+  outputOptions: {
+    comments: false,
+  },
 })


jerome-benoit:
The downsides are very little compared to the gains. The source maps are not generated during the build?

AriPerkkio:
I don't agree with that at all. Publishing developer tools as minified code is bad practice overall.

Including source maps in the published package would increase the package size even more. I would not recommend that either.

jerome-benoit:
> I don't agree with that at all. Publishing developer tools as minified code is bad practice overall.

It depends on the code purpose. The API is compatible with browser to allow direct usage. So code minified can stay for browser for example. Since the code is transpiled, the debugging to the exact original code can only be achieved with a source maps file. Adding a source maps allow the debugging and keep the "tiny" expectation given the github org purpose.    

> Including source maps in the published package would increase the package size even more. I would not recommend that either.

The package size is not the issue here, the issue is the effective code file size that ends up loaded by the JS runtime and that is meant to stay tiny. The source maps are only loaded on demand per user request.

43081j:
My opinion:

I'd prefer we do not turn source maps on as it is just a good way to bloat a package only a minority will ever want to debug. 

As for this change - as a library, I agree we shouldn't minify. For browsers, it does make some sense. 

Personally, I would just turn minification off assuming it doesn't make a huge difference to the size.

I guess the important question: how much bigger is it?

AriPerkkio:
```diff
- ℹ dist/index.js    32.79 kB │ gzip: 13.51 kB
+ ℹ dist/index.js    46.22 kB │ gzip: 15.17 kB
  ℹ dist/index.d.ts  22.93 kB │ gzip:  5.26 kB
- ℹ 2 files, total: 55.72 kB
+ ℹ 2 files, total: 69.15 kB
```

jerome-benoit:
> As for this change - as a library, I agree we shouldn't minify. For browsers, it does make some sense.

I guess tsdown can be configured to minify a target (.min.js) and not the others. And the package.json export the right target for browser, ESM or CJS ?

43081j:
it means we'd have to ship two bundles, which is out of the question i think.

so we really just need to decide one way or the other: minify or don't minify.

@jerome-benoit the extra 10KB is acceptable to make the code debuggable without shipping sourcemaps i think.

though the extra size is basically comments (which are not really useful since our DTS contains the same JSDoc).

we could also use this tsdown config it seems:

```ts
  outputOptions: {
    comments: false
  },
  minify: {
    compress: true,
    mangle: true,
    codegen: {
      removeWhitespace: false
    }
  },

```

so we strip comments and keep whitespace. then its 38KB vs 32KB iirc

AriPerkkio:
> It depends on the code purpose. The API is compatible with browser to allow direct usage. So code minified can stay for browser for example.

I think this decision should be based on Tinybench's users.

97.7% of Tinybench users are using it via Vitest. Maybe the remaining users are using it directly in browser. (Actually I'm quite sure most, maybe even 95% of that, is not even using Tinybench when they install Vitest, it's just transitive dependency for them)



- https://majors.nullvoxpopuli.com/q?minors=true&old=false&packages=tinybench

PR #563, opened 2026-05-21
chore: enable staged publishing

43081j:
Also bumps all the actions using a sha rather than a mutable tag.

### DIFF
--- a/.github/workflows/cr.yml
+++ b/.github/workflows/cr.yml
@@ -10,14 +10,14 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: true
 
       - name: Install pnpm
-        uses: pnpm/action-setup@v6
+        uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6.0.8
 
-      - uses: actions/setup-node@v6
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version: 24
           cache: pnpm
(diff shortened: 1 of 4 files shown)


jerome-benoit:
pnpm is not supporting staged publishing ?

43081j:
it was very recently added to npm (last day or two) so hasn't yet reached pnpm afaik.

though keep in mind `pnpm publish` does seem to do the same as `npm publish` in non-monorepo situations. so we're not losing anything by switching the publish command and can always flip it back later.

jerome-benoit:
Pushed `2d9cbff` switching the four `npm stage publish` calls back to `pnpm stage publish`.

`pnpm stage` shipped natively in pnpm v11.3.0 ([pnpm/pnpm#11863](https://github.com/pnpm/pnpm/pull/11863), released May 24) with full subcommand parity (`publish`, `list`, `view`, `approve`, `reject`, `download`). It reuses `libnpmpublish ≥ 11.2.0` so it talks to the same registry endpoint as `npm stage publish`. Flags (`--no-git-checks`, `--tag`, `--access`, `--provenance`, `--otp`, `--dry-run`, `--recursive`) are identical — see [pnpm.io/cli/stage](https://pnpm.io/cli/stage).

Since `package.json` is already pinned to `pnpm@11.3.0`, no version bump was needed. The two-step `npm install -g npm@~11.10.0 && npm install -g npm@11.15.0` workaround for [npm/cli#9151](https://github.com/npm/cli/issues/9151) is also dropped — only pnpm is needed in the publish job now. Bonus: `pnpm stage publish` rewrites `workspace:*` correctly, which `npm stage publish` does not (the original motivation for [pnpm/pnpm#11863](https://github.com/pnpm/pnpm/pull/11863), per @JoviDeCroock); future-proof if the project ever moves to a workspace layout.

The [e18e/setup-publish](https://github.com/e18e/setup-publish/blob/main/templates/default%2Bpnpm.yml) reference template (same ecosystem as tinylibs) already uses `pnpm stage publish`.

jerome-benoit (approved):

PR #568, opened 2026-05-30
feat: add opt-in subtractTimerOverhead option

jerome-benoit:
## Summary

Add a new `BenchOptions.subtractTimerOverhead` boolean (default `false`). When enabled, the cost of one timestamp provider call is calibrated once at construction time via the new exported `calibrateTimerOverhead` helper, and subtracted from each raw latency sample (clamped to zero) before statistics are computed.

The estimate is exposed as `bench.timerOverhead` (`number | undefined`).

## Implementation

- `calibrateTimerOverhead(provider, samples = 1024)` returns the median of the strictly positive deltas of back-to-back `now()` calls, or `0` on coarse-timer runtimes.
- The correction is applied **post-loop**, in `Task#processRunResult`, just before `sortSamples`. The sampling loop's exit condition keeps using raw timings and converges naturally even when the function under test runs at the timer grain — applying the correction inside `#measure`/`#measureSync` would zero-out samples on sub-µs functions and cause the OR-loop's `totalTime < time` condition to never be met.
- `overriddenDuration` samples are still collected via the same path and corrected uniformly; the correction is opt-in so a benchmark relying on overridden durations should leave it disabled.

## Caveats (documented in JSDoc)

Only location statistics (mean, percentiles) are corrected. Variance, sd, sem, moe and rme are unchanged because subtracting a constant does not reduce dispersion (`Var(X − C) = Var(X)` for a constant `C`). `bench.timerOverhead` is `undefined` when the option is off, `0` on runtimes whose timer resolution exceeds its own call cost.

## Risk

None for existing benchmarks — option defaults to `false`, behavior strictly unchanged when omitted. No new runtime dependency. Bundle size impact is minimal (calibration helper + 5 LOC in the hot path).

### DIFF
--- a/src/bench.ts
+++ b/src/bench.ts
@@ -24,6 +24,7 @@ import { BenchEvent } from './event'
 import { Task } from './task'
 import {
   assert,
+  calibrateTimerOverhead,
   defaultConvertTaskResultForConsoleTable,
   getTimestampProvider,
   runtime,
@@ -96,6 +97,13 @@ export class Bench extends EventTarget implements BenchLike {
   readonly signal?: AbortSignal
 
   /**
+   * Whether to subtract an estimated timestamp provider call overhead from
+   * each raw latency sample.
+   * @default false
+   */
+  readonly subtractTimerOverhead: boolean
+
+  /**
    * A teardown function that runs after each task execution.
    */
   readonly teardown: (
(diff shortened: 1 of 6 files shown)


jerome-benoit:
Closed in favor of #571 which consolidates this feature with #569 and #570 into a single coherent change. The 8-agent cross-validated audit identified that naively composing the three PRs would have caused the diagnostics in #569 and #570 to operate on the corrected-and-clamped sample set, producing artificially small detected-resolution values and false-positive saturation warnings whenever overhead subtraction was active. The consolidated PR also folds in the fixes for the issues surfaced by the same audit: bigint-precision-preserving subtraction in calibration (`toMs(b - a)`), JIT warmup before sample collection, coarse-timer fail-mode guard (returns `0` when `C < R/2`), `'min' | 'p05'` estimator alternatives, and `overriddenDuration`-aware correction skip.

PR #571, opened 2026-05-30
feat: add timer-overhead correction, saturation warning, and resolution diagnostic

jerome-benoit:
## Summary

Three cooperating timer diagnostics for sub-microsecond benchmarking, integrated in a single coherent code path:

1. **`subtractTimerOverhead`** — opt-in calibration of one timestamp provider call cost `Ĉ`, applied as `max(0, raw - Ĉ)` to non-overridden samples before statistics.
2. **`'warning'` event** — dispatched on both `Bench` and `Task` when a `Task`'s measured samples are dominated by the timer resolution. Carries a `TimerSaturationReason` payload (`'zero-dominated' | 'low-distinct' | 'zero-mad'`).
3. **`Task.detectedResolution`** — populated after each run with the smallest reproducibly observed positive sample among timer-measured samples.

Replaces #568, #569 and #570: their naïve composition would have the diagnostics operate on the corrected-and-clamped sample set, producing artificially small resolution estimates and false-positive saturation warnings.

## Public API

```ts
import { Bench } from 'tinybench'

// Per-Bench (opt-in, default false)
const bench = new Bench({ subtractTimerOverhead: true })
bench.subtractTimerOverhead  // readonly boolean
bench.timerOverhead          // number | undefined — calibrated Ĉ in ms

// Per-Task diagnostic
const task = bench.getTask('foo')
task.detectedResolution      // number | undefined

// Saturation warning event with typed reason
bench.addEventListener('warning', evt => {
  evt.task    // Task
  evt.reason  // TimerSaturationReason | undefined
})

// Standalone helpers
import {
  calibrateTimerOverhead,
  classifyTimerSaturation,    // → TimerSaturationReason | undefined
  detectTimerSaturation,      // boolean wrapper around classifyTimerSaturation
  estimateResolution,
  medianAbsoluteDeviation,    // mad statistic from a sorted sample
} from 'tinybench'

// Timestamp provider objects (pass to calibrateTimerOverhead)
import {
  hrtimeNowTimestampProvider,
  performanceNowTimestampProvider,
} from 'tinybench'

import type {
  CalibrateTimerOverheadOptions,
  TimerOverheadEstimatorKind,  // 'median' | 'min' | 'p05'
  TimerSaturationReason,
} from 'tinybench'
```

## `Task#processRunResult` ordering (cross-cutting fix)

`isOverridden[]` is collected in lockstep with `latencySamples[]` during measurement. The phase ordering preserves the alignment invariant and isolates user-supplied `overriddenDuration` values from timer diagnostics:

1. **Apply overhead correction** in-place on the collection-order array (skipping overridden indices). Alignment intact.
2. **Build a `measuredOnly` view** by filtering out overridden indices, before any sort.
3. **Sort** the working `latencySamples` array.
4. **`computeStatistics`** on the sorted (possibly corrected) samples.
5. **`estimateResolution` + `classifyTimerSaturation`** on the sorted measured-only subset (`estimateResolution` requires ascending-sorted input). With no overrides the subset is `latencySamples` itself; otherwise the filtered view is sorted first. An all-overridden run yields `undefined` resolution and no warning.
6. **Dispatch** the `'warning'` event (carrying the reason, when a criterion fired), then `'cycle'` and `'complete'`.

`isOverridden` is allocated unconditionally so the measured-only filter is also active when `subtractTimerOverhead: false` (covers `overriddenDuration` users without overhead correction).

## Constraints

- `subtractTimerOverhead` + `concurrency: 'task'` is rejected. Asserted at construction **and** re-checked at the start of `Bench.run()` to guard against untyped (JS-side) mutation of the `readonly` `concurrency` field after construction.
- When the timer is too coarse to resolve the call cost — fewer than half the calibration pairs produce a positive delta (`C < R / 2`, e.g. a `Date.now`-class timer with `>= 1 ms` resolution) — calibration returns `0` and the option becomes a no-op.

## Risk

- **Runtime:** opt-in for behaviour-changing pieces. Default-options benchmarks behave identically.
- **TypeScript:** `BenchEvents` widened with `'warning'`; exhaustive `switch` consumers under `noFallthroughCasesInSwitch` need a `case 'warning':` arm. `BenchLike.timerOverhead` is now `readonly` and optional. New types `TimerSaturationReason`, `TimerOverheadEstimatorKind`, `CalibrateTimerOverheadOptions`. New runtime exports `calibrateTimerOverhead`, `classifyTimerSaturation`, `detectTimerSaturation`, `estimateResolution`, `medianAbsoluteDeviation`, `hrtimeNowTimestampProvider`, `performanceNowTimestampProvider`.
- **Bundle:** `+0.7 kB` gzipped vs main. The pre-existing `size-limit` budget of 12 kB is already exceeded on main (14.15 kB) — this PR does not change that situation.
- **No new runtime dependency.**

## JSDoc honesty

The `subtractTimerOverhead` JSDoc describes the `max(0, x)` clamp explicitly:

- Two regimes (clean-shift `X >> Ĉ` vs sub-overhead `X ≈ Ĉ`)
- `rme` deterministically inflates by factor `M / (M − Ĉ)` whenever `Ĉ > 0`
- In the sub-overhead regime, `p50`, `mad`, `aad` collapse to `0` once cumulative mass at `≤ Ĉ` reaches the relevant quantile
- Three observable consequences of the clamp (`latency.min == 0`, throughput-mean substitution, `'zero-dominated'` criterion ambiguity)

## Verification

- `pnpm typecheck` clean
- `pnpm lint` clean
- `pnpm test` — **237 tests pass** across **64 test files**
- `pnpm build` — `dist` generated cleanly

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,10 +340,11 @@ const overhead = calibrateTimerOverhead(hrtimeNowTimestampProvider, {
 ## Per-Sample Override (`overriddenDuration`)
 
 A task function may return an object containing `overriddenDuration`
-(in ms). That value replaces the timer-measured sample directly,
-bypassing both the timer and any overhead correction. Useful for
-externally-timed work or sub-overhead measurements that the timer
-cannot resolve.
+(in ms). That value is recorded in place of the timer-measured sample:
+the timer still brackets the task function, but its measurement is
+discarded and overhead correction is not applied to the substituted
+value. Useful for externally-timed work or sub-overhead measurements
+that the timer cannot resolve.
 
 ```ts
 bench.add('externally-timed', () => {
--- a/README.md
+++ b/README.md
@@ -317,8 +317,8 @@ import { calibrateTimerOverhead, hrtimeNowTimestampProvider } from 'tinybench'
 
 const overhead = calibrateTimerOverhead(hrtimeNowTimestampProvider, {
   estimator: 'p05',
-  samples: 1024,
-  warmupSamples: 64,
+  pairs: 1024,
+  warmupPairs: 64,
 })
 ```
 
--- a/README.md
+++ b/README.md
@@ -330,7 +330,9 @@ const overhead = calibrateTimerOverhead(hrtimeNowTimestampProvider, {
 - For sub-overhead measurements (`X ≈ Ĉ`) the `max(0, …)` clamp
   truncates the lower tail and biases statistics; prefer
   `overriddenDuration` (see below).
-- On runtimes with a coarse timer (resolution >= 1 ms) the calibration
+- When the timer is too coarse to resolve the call cost — fewer than half
+  of the calibration pairs produce a positive delta (call cost `C < R / 2`,
+  e.g. a `Date.now`-class timer with `>= 1 ms` resolution) — the calibration
   returns `0` and the option becomes a no-op.
 
 ## Per-Sample Override (`overriddenDuration`)
--- a/README.md
+++ b/README.md
@@ -153,8 +153,10 @@ await bench.run()
 - When `mode` is set to 'bench', different tasks within the bench run concurrently. Concurrent cycles.
 
 ```ts
-bench.threshold = 10 // The maximum number of concurrent tasks to run. Defaults to Number.POSITIVE_INFINITY.
-bench.concurrency = 'task' // The concurrency mode to determine how tasks are run.
+const bench = new Bench({
+  concurrency: 'task', // The concurrency mode to determine how tasks are run.
+  threshold: 10, // The maximum number of concurrent tasks to run. Defaults to Number.POSITIVE_INFINITY.
+})
 await bench.run()
 ```
 
--- a/README.md
+++ b/README.md
@@ -87,6 +87,12 @@ Both the `Task` and `Bench` classes extend the `EventTarget` object. So you can 
 bench.addEventListener('cycle', (evt) => {
   const task = evt.task!;
 });
+
+// runs when timer saturation is detected for a task's measured samples
+bench.addEventListener('warning', (evt) => {
+  const task = evt.task!;
+  const reason = evt.reason; // 'zero-dominated' | 'low-distinct' | 'zero-mad'
+});
 ```
 
 #### [`TaskEvents`](https://tinylibs.github.io/tinybench/types/TaskEvents.html)
@@ -285,6 +291,96 @@ const bench = new Bench({
   now: Date.now,
 })
 ```
+
+## Timer Overhead Correction
+
+Each timer call (`performance.now()`, `process.hrtime.bigint()`, …) has a
+non-zero call cost `C`. For a task whose true duration `X` is comparable
+to `C`, the raw measured sample `X + C` is dominated by the timer rather
+than the task.
+
+When `subtractTimerOverhead: true` is set, an estimate `Ĉ` is computed
+once at construction time via [`calibrateTimerOverhead`](https://tinylibs.github.io/tinybench/functions/calibrateTimerOverhead.html),
+and `Math.max(0, raw_sample - Ĉ)` is used as each non-overridden sample
+before statistics are computed.
+
+```ts
+const bench = new Bench({ subtractTimerOverhead: true })
+console.log(bench.timerOverhead) // calibrated Ĉ in ms (or undefined)
+```
+
+The calibration helper is also exported for direct use, with a
+configurable estimator strategy (`'median'` default, or `'min'` / `'p05'`):
+
+```ts
+import { calibrateTimerOverhead, hrtimeNowTimestampProvider } from 'tinybench'
+
+const overhead = calibrateTimerOverhead(hrtimeNowTimestampProvider, {
+  estimator: 'p05',
+  samples: 1024,
+  warmupSamples: 64,
+})
+```
+
+**Caveats.**
+
+- Incompatible with `concurrency: 'task'` — overhead is calibrated
+  sequentially and does not reflect concurrent execution cost.
+  Construction (and `run()`) throws if both are set.
+- For sub-overhead measurements (`X ≈ Ĉ`) the `max(0, …)` clamp
+  truncates the lower tail and biases statistics; prefer
+  `overriddenDuration` (see below).
+- On runtimes with a coarse timer (resolution >= 1 ms) the calibration
+  returns `0` and the option becomes a no-op.
+
+## Per-Sample Override (`overriddenDuration`)
+
+A task function may return an object containing `overriddenDuration`
+(in ms). That value replaces the timer-measured sample directly,
+bypassing both the timer and any overhead correction. Useful for
+externally-timed work or sub-overhead measurements that the timer
+cannot resolve.
+
+```ts
+bench.add('externally-timed', () => {
+  const start = process.hrtime.bigint()
+  doWork()
+  const elapsedMs = Number(process.hrtime.bigint() - start) / 1e6
+  return { overriddenDuration: elapsedMs }
+})
+```
+
+Overridden samples are excluded from `Task.detectedResolution` and
+from timer-saturation detection.
+
+## Timer Diagnostics
+
+After `bench.run()` (or `runSync()`), each task exposes
+`detectedResolution` — the smallest reproducibly observed positive
+sample (in ms) among the timer-measured samples, or `undefined` when no
+positive timer measurement was seen (e.g. every sample was overridden).
+
+```ts
+const task = bench.getTask('foo')
+console.log(task?.detectedResolution) // e.g. 0.000041 (≈ 41 ns)
+```
+
+When the timer's resolution dominates a task's measured distribution
+(more than half zero samples, fewer than `max(3, min(10, ⌊n / 1000⌋))`
+distinct values, or zero MAD with `n > 100`), tinybench dispatches a
+`'warning'` event on both the task and the bench, carrying the matching
+[`TimerSaturationReason`](https://tinylibs.github.io/tinybench/types/TimerSaturationReason.html):
+
+```ts
+bench.addEventListener('warning', evt => {
+  console.warn(`timer-saturated: ${evt.task?.name} — ${evt.reason}`)
+})
+```
+
+The same heuristic and estimator are exposed as standalone helpers for
+custom analysis: [`detectTimerSaturation`](https://tinylibs.github.io/tinybench/functions/detectTimerSaturation.html),
+[`classifyTimerSaturation`](https://tinylibs.github.io/tinybench/functions/classifyTimerSaturation.html),
+and [`estimateResolution`](https://tinylibs.github.io/tinybench/functions/estimateResolution.html).
 
 ## Aborting Benchmarks
 
--- a/src/bench.ts
+++ b/src/bench.ts
@@ -31,6 +31,9 @@ import {
   runtimeVersion,
 } from './utils'
 
+const subtractTimerOverheadConcurrencyError =
+  '`subtractTimerOverhead` cannot be used with `concurrency: "task"` — set `concurrency` to `null` or `"bench"`, or disable `subtractTimerOverhead`'
+
 /**
  * The Bench class keeps track of the benchmark tasks and controls them.
  */
@@ -218,7 +221,7 @@ export class Bench extends EventTarget implements BenchLike {
     this.subtractTimerOverhead = restOptions.subtractTimerOverhead === true
     assert(
       !(this.subtractTimerOverhead && this.concurrency === 'task'),
-      '`subtractTimerOverhead` cannot be used with `concurrency: "task"` — set `concurrency` to `null` or `"bench"`, or disable `subtractTimerOverhead`'
+      subtractTimerOverheadConcurrencyError
     )
     this.timerOverhead = this.subtractTimerOverhead
       ? calibrateTimerOverhead(this.timestampProvider)
@@ -294,7 +297,7 @@ export class Bench extends EventTarget implements BenchLike {
   async run (): Promise<Task[]> {
     assert(
       !(this.subtractTimerOverhead && this.concurrency === 'task'),
-      '`subtractTimerOverhead` cannot be used with `concurrency: "task"` — set `concurrency` to `null` or `"bench"`, or disable `subtractTimerOverhead`'
+      subtractTimerOverheadConcurrencyError
     )
     if (this.warmup) {
       await this.#warmupTasks()
--- a/src/bench.ts
+++ b/src/bench.ts
@@ -99,6 +99,9 @@ export class Bench extends EventTarget implements BenchLike {
   /**
    * Whether to subtract an estimated timestamp provider call overhead from
    * each raw latency sample.
+   *
+   * Incompatible with `concurrency: 'task'`; the constraint is enforced
+   * at construction and at the start of {@link Bench.run}.
    * @default false
    */
   readonly subtractTimerOverhead: boolean
@@ -212,10 +215,10 @@ export class Bench extends EventTarget implements BenchLike {
     this.throws = restOptions.throws ?? false
     this.signal = restOptions.signal
     this.retainSamples = restOptions.retainSamples === true
-    this.subtractTimerOverhead = restOptions.subtractTimerOverhead ?? false
+    this.subtractTimerOverhead = restOptions.subtractTimerOverhead === true
     assert(
       !(this.subtractTimerOverhead && this.concurrency === 'task'),
-      '`subtractTimerOverhead` is incompatible with `concurrency: "task"` — overhead is calibrated sequentially and does not reflect concurrent execution cost'
+      '`subtractTimerOverhead` cannot be used with `concurrency: "task"` — set `concurrency` to `null` or `"bench"`, or disable `subtractTimerOverhead`'
     )
     this.timerOverhead = this.subtractTimerOverhead
       ? calibrateTimerOverhead(this.timestampProvider)
@@ -289,6 +292,10 @@ export class Bench extends EventTarget implements BenchLike {
    * @returns the tasks array
    */
   async run (): Promise<Task[]> {
+    assert(
+      !(this.subtractTimerOverhead && this.concurrency === 'task'),
+      '`subtractTimerOverhead` cannot be used with `concurrency: "task"` — set `concurrency` to `null` or `"bench"`, or disable `subtractTimerOverhead`'
+    )
     if (this.warmup) {
       await this.#warmupTasks()
     }
--- a/src/bench.ts
+++ b/src/bench.ts
@@ -212,7 +212,11 @@ export class Bench extends EventTarget implements BenchLike {
     this.throws = restOptions.throws ?? false
     this.signal = restOptions.signal
     this.retainSamples = restOptions.retainSamples === true
-    this.subtractTimerOverhead = restOptions.subtractTimerOverhead === true
+    this.subtractTimerOverhead = restOptions.subtractTimerOverhead ?? false
+    assert(
+      !(this.subtractTimerOverhead && this.concurrency === 'task'),
+      '`subtractTimerOverhead` is incompatible with `concurrency: "task"` — overhead is calibrated sequentially and does not reflect concurrent execution cost'
+    )
     this.timerOverhead = this.subtractTimerOverhead
       ? calibrateTimerOverhead(this.timestampProvider)
       : undefined
--- a/src/bench.ts
+++ b/src/bench.ts
@@ -103,8 +103,10 @@ export class Bench extends EventTarget implements BenchLike {
    * Whether to subtract an estimated timestamp provider call overhead from
    * each raw latency sample.
    *
-   * Incompatible with `concurrency: 'task'`; the constraint is enforced
-   * at construction and at the start of {@link Bench.run}.
+   * Incompatible with `concurrency: 'task'`. Enforced at construction and
+   * re-checked at the start of {@link Bench.run} to guard against untyped
+   * (JS-side) mutation of the `readonly` `concurrency` field after
+   * construction.
    * @default false
    */
   readonly subtractTimerOverhead: boolean
--- a/src/bench.ts
+++ b/src/bench.ts
@@ -24,6 +24,7 @@ import { BenchEvent } from './event'
 import { Task } from './task'
 import {
   assert,
+  calibrateTimerOverhead,
   defaultConvertTaskResultForConsoleTable,
   getTimestampProvider,
   runtime,
@@ -96,6 +97,13 @@ export class Bench extends EventTarget implements BenchLike {
   readonly signal?: AbortSignal
 
   /**
+   * Whether to subtract an estimated timestamp provider call overhead from
+   * each raw latency sample.
+   * @default false
+   */
+  readonly subtractTimerOverhead: boolean
+
+  /**
    * A teardown function that runs after each task execution.
    */
   readonly teardown: (
@@ -119,6 +127,15 @@ export class Bench extends EventTarget implements BenchLike {
    * The amount of time to run each task.
    */
   readonly time: number
+
+  /**
+   * The estimated cost of one timestamp provider call in milliseconds.
+   *
+   * `undefined` when {@link subtractTimerOverhead} is `false`.
+   * Otherwise calibrated once at construction time via
+   * {@link calibrateTimerOverhead}.
+   */
+  readonly timerOverhead: number | undefined
 
   /**
    * A timestamp provider and its related functions.
@@ -195,6 +212,10 @@ export class Bench extends EventTarget implements BenchLike {
     this.throws = restOptions.throws ?? false
     this.signal = restOptions.signal
     this.retainSamples = restOptions.retainSamples === true
+    this.subtractTimerOverhead = restOptions.subtractTimerOverhead === true
+    this.timerOverhead = this.subtractTimerOverhead
+      ? calibrateTimerOverhead(this.timestampProvider)
+      : undefined
 
     if (this.signal) {
       this.signal.addEventListener(
--- a/src/index.ts
+++ b/src/index.ts
@@ -32,6 +32,7 @@ export type {
   TaskResultStarted,
   TaskResultTimestampProviderInfo,
   TaskResultWithStatistics,
+  TimerSaturationReason,
   TimestampFn,
   TimestampFns,
   TimestampProvider,
@@ -40,7 +41,6 @@ export type {
 export type {
   CalibrateTimerOverheadOptions,
   TimerOverheadEstimatorKind,
-  TimerSaturationReason,
 } from './utils'
 export {
   calibrateTimerOverhead,
--- a/src/index.ts
+++ b/src/index.ts
@@ -49,8 +49,10 @@ export {
   estimateResolution,
   formatNumber,
   hrtimeNow,
+  hrtimeNowTimestampProvider,
   medianAbsoluteDeviation,
   mToNs,
   performanceNow as now,
   nToMs,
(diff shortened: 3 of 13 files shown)


src/task.ts
```diff
@@ -555,26 +585,84 @@ export class Task extends EventTarget {
...
+      const measuredOnly: number[] = hasAnyOverridden
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        ? latencySamples.filter((_, i) => isOverridden![i] !== true)
+        : latencySamples
+
+      // Phase 3 — Single sort of the working array.
       sortSamples(latencySamples)
 
+      // Phase 4 — Resolution diagnostic on sorted samples.
+      this.#detectedResolution = estimateResolution(latencySamples)
```
jerome-benoit: `estimateResolution(latencySamples)` runs on the working array, which still contains `overriddenDuration` values when any iteration overrode the timer. A task returning a constant `overriddenDuration` reports that constant as `detectedResolution`. For consistency with Phase 6, pass `measuredOnly` here, or amend the field's docstring to state it reflects all samples (timer + overridden) rather than the timer.

src/task.ts
```diff
@@ -555,26 +585,84 @@ export class Task extends EventTarget {
...
+      const overhead = this.#bench.timerOverhead
+      const hasOverhead = overhead !== undefined && overhead > 0
+
+      // Phase 1 — Subtract overhead while isOverridden[i] is still aligned with
+      // latencySamples[i] (both in collection order, pre-sort).
+      if (hasOverhead) {
+        for (let i = 0; i < latencySamples.length; i++) {
+          if (isOverridden?.[i] !== true) {
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            latencySamples[i] = Math.max(0, latencySamples[i]! - overhead)
```
jerome-benoit, line 640: `Math.max(0, sample - overhead)` collapses every raw sample ≤ `overhead` to exactly 0. Three downstream consequences: `latency.min` is reported as 0 with no real zero-duration run; the throughput zero-branch (line 675) fires more often; criterion A of `detectTimerSaturation` (`zeroCount * 2 > n`) can fire from clamping rather than genuine saturation. Worth at minimum a doc note on `subtractTimerOverhead`.

src/task.ts
```diff
@@ -555,26 +585,84 @@ export class Task extends EventTarget {
...
         this.#retainSamples
       )
+
+      // Phase 6 — Saturation detection on measured-only samples.
+      let saturated = false
+      if (measuredOnly === latencySamples) {
+        saturated = detectTimerSaturation(latencySamples, latencyStatistics.mad)
+      } else if (isValidSamples(measuredOnly)) {
+        sortSamples(measuredOnly)
+        const measuredStats = computeStatistics(measuredOnly, false)
```
jerome-benoit: `computeStatistics(measuredOnly, false)` is invoked solely to read `.mad`; the other 14 fields (mean, vr, sd, sem, moe, rme, aad, p50, p75, p99, p995, p999, min, max, samplesCount) are computed and discarded. Replace with `absoluteDeviationMedian(measuredOnly, quantileSorted(measuredOnly, 0.5))` to skip the wasted work on the mixed-override path.

src/task.ts
```diff
@@ -606,6 +694,12 @@ export class Task extends EventTarget {
         totalTime,
       }
       /* eslint-enable perfectionist/sort-objects */
+
+      if (saturated) {
+        const warningEv = new BenchEvent('warning', this)
```
jerome-benoit: The `'warning'` event exposes only `task`. `detectTimerSaturation` aggregates three criteria (zero-dominated, low-distinct-count, zero-MAD with `n > 100`) into one boolean; consumers cannot differentiate them without re-running detection (which requires `retainSamples: true`). Return a reason from `detectTimerSaturation` and thread it onto the event detail.

src/bench.ts
```diff
@@ -195,6 +212,14 @@ export class Bench extends EventTarget implements BenchLike {
     this.throws = restOptions.throws ?? false
     this.signal = restOptions.signal
     this.retainSamples = restOptions.retainSamples === true
+    this.subtractTimerOverhead = restOptions.subtractTimerOverhead ?? false
```
jerome-benoit: `?? false` accepts any non-`undefined` truthy value from JS callers and stores it verbatim in a `boolean`-typed field. The line above (`retainSamples = restOptions.retainSamples === true`) coerces non-`true` to `false`. Match the sibling style: `restOptions.subtractTimerOverhead === true`.

src/bench.ts
```diff
@@ -195,6 +212,14 @@ export class Bench extends EventTarget implements BenchLike {
...
+    assert(
+      !(this.subtractTimerOverhead && this.concurrency === 'task'),
+      '`subtractTimerOverhead` is incompatible with `concurrency: "task"` — overhead is calibrated sequentially and does not reflect concurrent execution cost'
+    )
```
jerome-benoit, line 227: The assert runs once at construction. `concurrency` is `readonly` only at the TS level; the README documents the post-construction mutation pattern `bench.concurrency = 'task'` before `await bench.run()`. A user enabling `subtractTimerOverhead: true` then mutating `concurrency` to `'task'` post-construction silently runs the disallowed combination with `timerOverhead` populated from sequential calibration. Move the check into `run()`/`runSync()` (matching the pattern at line 322) or freeze the field at runtime.

src/bench.ts
```diff
@@ -195,6 +212,14 @@ export class Bench extends EventTarget implements BenchLike {
     this.throws = restOptions.throws ?? false
     this.signal = restOptions.signal
     this.retainSamples = restOptions.retainSamples === true
+    this.subtractTimerOverhead = restOptions.subtractTimerOverhead ?? false
+    assert(
+      !(this.subtractTimerOverhead && this.concurrency === 'task'),
+      '`subtractTimerOverhead` is incompatible with `concurrency: "task"` — overhead is calibrated sequentially and does not reflect concurrent execution cost'
```
jerome-benoit: The message states the cause but not the remediation. Append: `` — set `concurrency` to `'bench'` or `null`, or disable `subtractTimerOverhead` ``.

src/types.ts
```diff
@@ -120,6 +121,11 @@ export interface BenchLike extends EventTarget {
    * The amount of time to run each task.
    */
   time: number
+  /**
+   * The estimated cost of one timestamp provider call in milliseconds, or
+   * `undefined` when timer overhead subtraction is disabled.
+   */
+  timerOverhead: number | undefined
```
jerome-benoit: `timerOverhead: number | undefined` is required on a public interface used for decoupling/dependency-injection. Third-party `BenchLike` implementers (test doubles, alternative bench harnesses) fail to type-check until they add the field. Make it optional (`timerOverhead?: number | undefined`) or call out the break in the changelog. The field is also missing the `readonly` modifier present on the class declaration (`bench.ts:138`); `Task` reads it on every cycle to gate correction, so allowing mid-run mutation through the interface is unsafe.

src/types.ts
```diff
@@ -183,6 +189,48 @@ export interface BenchOptions {
...
+   *
+   * Each sample is measured as `t1 - t0` around a single call to the task
+   * function, so every raw sample is inflated by approximately one
+   * timestamp provider call cost `C`. When this option is `true`, an
+   * estimate `Ĉ` is computed once at construction time via
+   * {@link calibrateTimerOverhead}, and `max(0, raw_sample - Ĉ)` is used
+   * in place of each non-overridden sample before statistics are computed.
+   *
+   * **Statistics after correction.** All statistics (mean, percentiles,
+   * variance, sd, sem, moe, rme) are computed on the clamped corrected
```
jerome-benoit: The list omits `aad`, `mad`, `min`, `max` from the stats affected by correction. `min` becomes 0 once any sample clamps; `mad` feeds the saturation criterion at line 305 of `utils.ts`. Replace the enumeration with "all fields of `Statistics`" or extend it.

src/types.ts
```diff
@@ -183,6 +189,48 @@ export interface BenchOptions {
...
+   * samples, not on the original distribution. When raw samples comfortably
+   * exceed `Ĉ` (`X >> Ĉ`), the clamp `max(0, …)` rarely triggers and the
+   * correction approximates a clean shift: location statistics improve and
+   * dispersion statistics are largely unaffected. When raw samples are
+   * comparable to the overhead (`X ≈ Ĉ`, typical for nano-scale
+   * operations), the clamp truncates the lower tail of the distribution,
+   * which introduces a small positive bias on the corrected mean and
+   * contracts variance/sd/sem/moe while potentially inflating rme. For
```
jerome-benoit: In the clean-shift regime (`X >> Ĉ`), corrected samples are an exact shift `xᵢ − Ĉ`, leaving variance/sd/sem/moe identical while the mean drops by `Ĉ`. Therefore `rme = moe / mean × 100` deterministically inflates by a factor of `M / (M − Ĉ)`, not "potentially". Move the rme remark out of the X ≈ Ĉ branch (where it compounds with sem contraction) and state both regimes inflate rme.

src/types.ts
```diff
@@ -183,6 +189,48 @@ export interface BenchOptions {
...
+   *
+   * **Statistics after correction.** All statistics (mean, percentiles,
+   * variance, sd, sem, moe, rme) are computed on the clamped corrected
+   * samples, not on the original distribution. When raw samples comfortably
+   * exceed `Ĉ` (`X >> Ĉ`), the clamp `max(0, …)` rarely triggers and the
+   * correction approximates a clean shift: location statistics improve and
+   * dispersion statistics are largely unaffected. When raw samples are
+   * comparable to the overhead (`X ≈ Ĉ`, typical for nano-scale
+   * operations), the clamp truncates the lower tail of the distribution,
+   * which introduces a small positive bias on the corrected mean and
```
jerome-benoit: When raw samples are comparable to `Ĉ`, more than half can clamp to 0, which forces `quantileSorted(samples, 0.5)` to return 0 and propagates into `mad` (and `aad`, throughput zero-branch). The doc only mentions a "small positive bias on the corrected mean"; the median is a location statistic that can move in the opposite direction and become structurally zero.

src/types.ts
```diff
@@ -23,6 +23,7 @@ export type BenchEvents =
   | 'reset' // when the reset method gets called
   | 'start' // when running the benchmarks gets started
   | 'warmup' // when the benchmarks start getting warmed up
+  | 'warning' // when timer saturation is detected for a task's latency samples
```
jerome-benoit, line 26: `'warning'` is added to `BenchEvents`, `BenchEventsWithTask` (line 45) and `TaskEvents` (line 589) simultaneously. Exhaustive `switch` consumers with `noFallthroughCasesInSwitch` get a TS error. Confirm the release will be a minor (not patch) bump and call this out in the changelog.

src/utils.ts
```diff
@@ -260,6 +260,90 @@ export const isValidSamples = (
...
+ *
+ * The distinct-value count is computed in O(n) by exploiting the
+ * sorted-ascending invariant of `samples` (`Task#processRunResult` calls
+ * `sortSamples` before this function).
+ * @param samples - the latency samples, sorted ascending
+ * @param mad - the median absolute deviation, as computed by computeStatistics
+ * @returns true when the timer resolution dominates the measurement
+ */
+export const detectTimerSaturation = (
+  samples: Samples,
```
jerome-benoit: The exported parameter type is `Samples`, but correctness requires sorted-ascending input — the O(n) distinct-count loop relies on the invariant. A misuse from outside on unsorted samples silently returns wrong results. Either change the type to `SortedSamples`, or sort internally on a copy.

src/utils.ts
```diff
@@ -260,6 +260,90 @@ export const isValidSamples = (
...
+  for (const s of samples) {
+    if (s === 0) zeroCount++
+  }
+  if (zeroCount * 2 > n) return true
+
+  let distinctCount = 1
+  for (let i = 1; i < n; i++) {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    if (samples[i]! !== samples[i - 1]!) distinctCount++
+  }
```
jerome-benoit, line 302: The distinct-count loop always traverses all `n − 1` pairs, even after `distinctCount` reaches `distinctThreshold`. Add `if (distinctCount >= distinctThreshold) break` once the threshold is met (criterion B already known to be false), or drop the implication of an exploited short-circuit from the JSDoc.

src/utils.ts
```diff
@@ -332,6 +416,110 @@ const quantileSorted = (
...
+ *
+ * - `'median'` — median of strictly-positive deltas (default). Robust to
+ *   occasional OS-scheduling jitter and GC spikes at the cost of a slight
+ *   upward bias on noisy hosts.
+ * - `'min'` — minimum of strictly-positive deltas. Captures the lowest
+ *   observed call cost.
+ * - `'p05'` — 5th percentile of strictly-positive deltas. A compromise
+ *   between robustness and tightness.
+ */
+export type TimerOverheadEstimator = 'median' | 'min' | 'p05'
```
jerome-benoit: `'median' | 'min' | 'p05'` is a string literal union. The `Estimator` suffix suggests a strategy/function pluggable by users; the option only accepts the three preset strings. Rename to `TimerOverheadEstimatorKind` or `TimerOverheadStrategy` to match what the value actually is.

src/utils.ts
```diff
@@ -332,6 +416,110 @@ const quantileSorted = (
...
+ */
+export const calibrateTimerOverhead = (
+  provider: TimestampProvider,
+  options: CalibrateTimerOverheadOptions = {}
+): number => {
+  const { estimator = 'median', samples = 1024, warmupSamples = 64 } = options
+  const { fn, toMs } = provider
+
+  for (let i = 0; i < warmupSamples; i++) {
+    const a = fn() as unknown as number
```
jerome-benoit: `fn() as unknown as number` on a provider whose `fn` returns `bigint` lies to the type system. At runtime, `b - a` runs as bigint subtraction (exact, ~10⁹ ns range) and `toMs` (e.g. `nBigintToMs`) accepts bigint, so behaviour is correct. The double cast defeats the type system: a future change to `toMs` signature would be unguarded. Type the locals as `TimestampValue` and rely on the duck-typed `−`.

test/subtract-timer-overhead-overridden.test.ts
jerome-benoit, line 1: All three tests in this file produce 100% overridden iterations, so `isOverridden` is uniformly `true` and the Phase 1/2 alignment invariant in `task.ts:625-642` is never exercised against a heterogeneous sample set — the case the alignment fix targets. Add a test where a counter-driven `fn` alternates `overriddenDuration` and measured iterations, asserting (a) overridden values land unchanged in `result.latency.samples` (with `retainSamples: true`), (b) measured values are reduced by `bench.timerOverhead` and clamped at 0, (c) statistics on the corrected mixed set are correct.

test/calibrate-timer-overhead.test.ts
```diff
@@ -0,0 +1,108 @@
...
+test('calibrateTimerOverhead estimator min is less than or equal to median', () => {
+  const med = calibrateTimerOverhead(hrtimeNowTimestampProvider, {
+    estimator: 'median',
+    samples: 512,
+  })
+  const min = calibrateTimerOverhead(hrtimeNowTimestampProvider, {
+    estimator: 'min',
+    samples: 512,
+  })
+  expect(min).toBeLessThanOrEqual(med * 2)
```
jerome-benoit: The title claims "min less than or equal to median" but the assertion is `min ≤ median * 2`. By definition `min ≤ median` for any non-empty distribution, so the looser bound passes for unrelated reasons. Tighten to `expect(min).toBeLessThanOrEqual(med)`.

test/calibrate-timer-overhead.test.ts
jerome-benoit, line 1: Two coverage gaps in this file:

1. The `'p05'` branch of `calibrateTimerOverhead` (`src/utils.ts:510-514`, including the `Math.max(0, Math.ceil(deltas.length * 0.05) - 1)` index math) has no test. Add a deterministic-provider call so the index formula is locked.
2. The constructor assert `subtractTimerOverhead && concurrency: 'task'` (`src/bench.ts:217-220`) has no regression test. Add `expect(() => new Bench({ subtractTimerOverhead: true, concurrency: 'task' })).toThrow(/incompatible/)`.

test/detected-resolution.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  if (!fooTask) return
+  expect(fooTask.result.state).toBe('completed')
+  if (fooTask.result.state !== 'completed') return
+
+  const resolution = fooTask.detectedResolution
+  if (resolution !== undefined) {
+    expect(resolution).toBeTypeOf('number')
+    expect(resolution).toBeGreaterThan(0)
+    expect(Number.isFinite(resolution)).toBe(true)
+  }
```
jerome-benoit: The `if (resolution !== undefined) { … }` block skips all assertions when `resolution` is undefined; a regression that always returns `undefined` (e.g. an `estimateResolution` short-circuit) passes silently. Replace the guard with `expect(resolution).toBeDefined()` followed by the existing assertions.

src/index.ts
```diff
@@ -37,4 +37,16 @@ export type {
...
+} from './utils'
+export {
+  calibrateTimerOverhead,
+  detectTimerSaturation,
+  estimateResolution,
+  formatNumber,
+  hrtimeNow,
+  performanceNow as now,
+  nToMs,
+} from './utils'
```
jerome-benoit, line 58: `subtractTimerOverhead`, the `'warning'` event, `Task.detectedResolution`, and the helpers `calibrateTimerOverhead` / `detectTimerSaturation` / `estimateResolution` are exported from `src/index.ts:38-52` but absent from the README. Users cannot discover any of this from the project's primary documentation. Add a section under "Docs" before "Aborting Benchmarks".

jerome-benoit (commented):
## Cross-validated audit (8 oracle agents, 4 paired dimensions)

Findings classified by inter-pair convergence:

**Convergent (≥2 pairs flagged the same item)**
- `task.ts:648` — `detectedResolution` includes `overriddenDuration` values
- `task.ts:662` — Phase 6 `computeStatistics` discards 14 of 15 fields
- `bench.ts:215` — `?? false` accepts truthy non-boolean from JS callers, inconsistent with sibling line 214
- `bench.ts:219` — assert is one-shot at construction; runtime `concurrency` mutation bypasses it
- `bench.ts:218` — assert message lacks remediation
- `task.ts:699` — `'warning'` event payload carries no reason
- `types.ts:128` — `BenchLike.timerOverhead` required + non-readonly: breaking + unsafe
- `types.ts:204` — JSDoc enumeration omits `aad`/`mad`/`min`/`max`
- `tests` — mixed overridden+measured run not exercised (the alignment invariant)
- `tests` — `concurrency:'task'` + `subtractTimerOverhead` assert untested
- `tests` — `'p05'` estimator branch untested
- `utils.ts:488` — bigint cast misleads the type system

**Single-agent / high-signal**
- `task.ts:631` — clamp injects synthetic zeros affecting `min`, throughput, criterion A
- `types.ts:212` — `rme` deterministically inflates by `M / (M − Ĉ)` in the clean-shift regime, not "potentially"
- `types.ts:211` — `p50` (and downstream `mad`/`aad`) can collapse to 0 when ≥ 50% samples clamp
- `types.ts:26` — `'warning'` widening across three unions breaks exhaustive `switch` consumers
- `utils.ts:285` — `detectTimerSaturation` accepts unsorted input but correctness requires sorted
- `utils.ts:301` — distinct-count loop has no short-circuit at threshold
- `utils.ts:454` — `TimerOverheadEstimator` is a string literal union; suffix suggests a strategy
- `test/calibrate-timer-overhead.test.ts:57` — `min ≤ median * 2` is mathematically trivial
- `test/detected-resolution.test.ts:59` — conditional `if (resolution !== undefined)` skips all assertions on regression
- `README.md` — public API additions undocumented

**Verified correct (sample)**
- Phase 1 alignment of `latencySamples[i]` with `isOverridden[i]` in collection order
- `measuredOnly === latencySamples` reference-equality optimization
- Aborted-run paths (no samples, partial samples)
- `estimateResolution` edge cases (`[0,0,0]` → `undefined`, all-equal positives)
- `'warning'` widening consistent across `BenchEvents` / `BenchEventsWithTask` / `TaskEvents`
- Bigint precision preserved at runtime
- Style and Biome conformance on new test files

Detail in the inline comments. The mixed-overridden test gap is the only blocker.

src/types.ts
jerome-benoit: Position taken (cross-validated by the design pair on `src/types.ts:26`): keep the widening — splitting `BenchEvents` into two unions would break the single-union contract that the rest of the API relies on (`addEventListener<K extends BenchEvents>`). To document in the changelog as a **minor** bump with explicit note: *"`BenchEvents`, `BenchEventsWithTask`, `TaskEvents` are widened with `'warning'`; exhaustive `switch` consumers under `noFallthroughCasesInSwitch` must add a `case 'warning':` arm"*. Final call on the bump (minor vs patch) is yours; if patch is preferred, we revert the widening and expose `'warning'` via a separate type union.

README.md
```diff
@@ -286,6 +292,96 @@ const bench = new Bench({
...
+```ts
+const bench = new Bench({ subtractTimerOverhead: true })
+console.log(bench.timerOverhead) // calibrated Ĉ in ms (or undefined)
+```
+
+The calibration helper is also exported for direct use, with a
+configurable estimator strategy (`'median'` default, or `'min'` / `'p05'`):
+
+```ts
+import { calibrateTimerOverhead, hrtimeNowTimestampProvider } from 'tinybench'
```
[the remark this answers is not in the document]
jerome-benoit: Fixed in 8600893 — `hrtimeNowTimestampProvider` and `performanceNowTimestampProvider` are now exported from `src/index.ts`, so the README example imports resolve.

src/task.ts
```diff
@@ -348,6 +374,7 @@ export class Task extends EventTarget {
 
       let totalTime = 0 // ms
       const samples: number[] = []
+      const isOverridden: boolean[] = []
```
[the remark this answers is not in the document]
jerome-benoit: Fixed in 15f30d9 — the parallel `boolean[]` is replaced by a `Set<number>` of overridden indices, populated only on override, so the all-measured path allocates nothing per sample. Alignment and concurrency are preserved (the index is taken synchronously right after the sample push).

README.md
```diff
@@ -286,6 +292,98 @@ const bench = new Bench({
...
+- When the timer is too coarse to resolve the call cost — fewer than half
+  of the calibration pairs produce a positive delta (call cost `C < R / 2`,
+  e.g. a `Date.now`-class timer with `>= 1 ms` resolution) — the calibration
+  returns `0` and the option becomes a no-op.
+
+## Per-Sample Override (`overriddenDuration`)
+
+A task function may return an object containing `overriddenDuration`
+(in ms). That value replaces the timer-measured sample directly,
+bypassing both the timer and any overhead correction. Useful for
```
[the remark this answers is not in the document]
jerome-benoit: Fixed in eca70bf — the README now states the timer still brackets the task function; only its measured value is discarded and overhead correction skipped.

src/utils.ts
```diff
@@ -332,6 +430,118 @@ const quantileSorted = (
...
+ */
+export const calibrateTimerOverhead = (
+  provider: TimestampProvider,
+  options: CalibrateTimerOverheadOptions = {}
+): number => {
+  const { estimator = 'median', pairs = 1024, warmupPairs = 64 } = options
+  const { fn, toMs } = provider
+
+  // Degenerate input: no pairs to measure ⇒ no overhead estimate.
+  if (pairs === 0) return 0
```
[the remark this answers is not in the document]
jerome-benoit: Fixed in f806350 — `calibrateTimerOverhead` rejects non-finite/non-integer/non-positive `pairs` (returns 0) and skips warmup for invalid `warmupPairs` via `Number.isInteger` guards, so `Infinity`/`NaN` no longer hang. Regression tests added.

src/task.ts
```diff
@@ -555,26 +587,95 @@ export class Task extends EventTarget {
...
+      const hasOverhead = overhead !== undefined && overhead > 0
+
+      // Phase 1 — Subtract overhead while isOverridden[i] is still aligned with
+      // latencySamples[i] (both in collection order, pre-sort).
+      if (hasOverhead) {
+        for (let i = 0; i < latencySamples.length; i++) {
+          if (isOverridden?.[i] !== true) {
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            latencySamples[i] = Math.max(0, latencySamples[i]! - overhead)
+          }
```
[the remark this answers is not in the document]
jerome-benoit: Intentional and documented — the `Task.detectedResolution` getter JSDoc states that under `subtractTimerOverhead` the value is derived from the overhead-corrected samples rather than the raw grain. Deferred as a possible future refinement.

src/utils.ts
```diff
@@ -260,6 +261,103 @@ export const isValidSamples = (
...
+export const estimateResolution = (
+  samples: SortedSamples
+): number | undefined => {
+  let fallbackMin: number | undefined
+  for (let i = 0; i < samples.length; i++) {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const v = samples[i]!
+    if (v <= 0) continue
+    fallbackMin ??= v
+    if (samples[i + 1] === v) return v
```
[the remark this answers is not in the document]
jerome-benoit: By design — `detectedResolution` is documented as the smallest reproducibly observed positive sample among timer-measured samples, not a probe of the provider grain. Deferred as a separate enhancement.

src/utils.ts
```diff
@@ -332,6 +430,114 @@ const quantileSorted = (
...
+    const b = fn() as bigint
+    toMs(b - a)
+  }
+
+  const deltas: number[] = []
+  for (let i = 0; i < samples; i++) {
+    const a = fn() as bigint
+    const b = fn() as bigint
+    const delta = toMs(b - a)
+    if (delta > 0) deltas.push(delta)
```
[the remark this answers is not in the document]
jerome-benoit: Documented best-effort — calibration returns 0 when fewer than half the pairs yield a positive delta (`C < R/2`); the coarse-timer no-op and the clamp bias are called out in the JSDoc and README. Deferred as a heuristic refinement.

src/task.ts
```diff
@@ -606,6 +707,12 @@ export class Task extends EventTarget {
         totalTime,
       }
       /* eslint-enable perfectionist/sort-objects */
+
+      if (saturationReason !== undefined) {
+        const warningEv = new BenchEvent('warning', this, saturationReason)
+        this.dispatchEvent(warningEv)
+        this.#bench.dispatchEvent(warningEv)
```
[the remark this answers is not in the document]
jerome-benoit: The `'warning'` dispatch reuses one event for the task and the bench, matching the existing `'cycle'`/`'error'` dispatch pattern in this file. If cross-target event reuse is a concern, it is pre-existing and repo-wide rather than specific to this PR.

src/bench.ts
```diff
@@ -264,6 +292,10 @@ export class Bench extends EventTarget implements BenchLike {
    * @returns the tasks array
    */
   async run (): Promise<Task[]> {
+    assert(
+      !(this.subtractTimerOverhead && this.concurrency === 'task'),
+      '`subtractTimerOverhead` cannot be used with `concurrency: "task"` — set `concurrency` to `null` or `"bench"`, or disable `subtractTimerOverhead`'
+    )
```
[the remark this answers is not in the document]
jerome-benoit: `concurrency` is `readonly`; only untyped/cast mutation can reach this path. The constructor and `run()` asserts cover the documented configuration; re-checking after every post-`start` mutation is out of scope.

PR #622, opened 2026-07-29
chore: add .gitattributes for consistent line endings

chernojagne:
## Summary

Adds a `.gitattributes` file to normalise line endings across platforms.

- `* text=auto` enables automatic text/binary detection and CRLF normalisation
- Binary markers for fonts, archives, node lockfiles (`pnpm-lock.yaml`, `yarn.lock`, `bun.lock`, `package-lock.json`), images, compiled objects, and media files
- `*.map text -diff` suppresses noisy diffs on source maps
- `*.patch` / `*.diff` marked `-text` to preserve byte sequences

The Windows CI matrix (Node 22/24/latest + Bun + Deno on `windows-latest`) makes consistent CRLF handling particularly relevant — this ensures contributors on Windows don't inadvertently commit CRLF line endings.

## AI usage

This change was generated by [Autar](https://autar.ai) and reviewed by a human before submission.


.gitattributes
```diff
@@ -0,0 +1,92 @@
...
+*.ogv binary
+*.webm binary
+*.wmv binary
+
+# Lockfiles
+package-lock.json binary
+yarn.lock binary
+pnpm-lock.yaml binary
+bun.lock binary
+
```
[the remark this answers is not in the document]
chernojagne: Good catch — you're right that `binary` disables text normalization entirely, which defeats the purpose here. Changed all four lockfile entries to `text eol=lf` in 56aa10c so CRLF normalization applies and line endings are forced to LF.

.gitattributes
```diff
@@ -0,0 +1,92 @@
+# Auto-detect text files and normalize line endings
+* text=auto
```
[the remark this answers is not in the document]
chernojagne: Good point — `* text=auto` without `eol=lf` leaves line-ending conversion up to each user's `core.autocrlf`, which is inconsistent. Since `.editorconfig` already standardizes on LF, I've added `eol=lf` to the global rule in 56aa10c to enforce that at the Git layer too.

.gitattributes
```diff
@@ -0,0 +1,92 @@
...
+*.png binary
+*.jpg binary
+*.jpeg binary
+*.gif binary
+*.tif binary
+*.tiff binary
+*.ico binary
+*.webp binary
+*.bmp binary
+*.psd binary
```
jerome-benoit: This entire binary catalogue (images/fonts/archives/objects/exe/audio/video/documents, lines 4–83) matches **zero** tracked files, and none are plausible for a tiny TS benchmarking lib. `text=auto` already auto-detects binaries by content — verified: a NUL-containing file with only `* text=auto` resolves to `i/-text w/-text`, identical to an explicit `binary` rule. Suggest dropping all of these blocks.

.gitattributes
```diff
@@ -0,0 +1,92 @@
...
+*.mpg binary
+*.ogv binary
+*.webm binary
+*.wmv binary
+
+# Lockfiles — text with forced LF so CRLF is never committed
+package-lock.json text eol=lf
+yarn.lock text eol=lf
+pnpm-lock.yaml text eol=lf
+bun.lock text eol=lf
```
jerome-benoit: Redundant with `* text=auto eol=lf` above: for always-text files, `text` (Set) and `text=auto` yield the identical `eol: lf` result. This is also a pnpm-only repo — `package-lock.json`, `yarn.lock`, `bun.lock` match nothing. Suggest removing the block; `pnpm-lock.yaml` is already covered by the wildcard. Minor: `text` (Set) is marginally less safe than `auto`, since it would force EOL conversion even on binary-looking content.

.gitattributes
```diff
@@ -0,0 +1,92 @@
...
+
+# Source maps — suppress noisy diffs (Web.gitattributes)
```
jerome-benoit: `(Web.gitattributes)` is a leftover provenance note from the upstream template. Also, no `.map` files are tracked (`dist/` is gitignored), so `*.map text -diff` is inert here. Suggest dropping both.

jerome-benoit (changes requested):
Thanks for tackling cross-platform line endings — the intent is right and the Windows CI matrix makes it relevant.

The core is correct: `* text=auto eol=lf` alone already does 100% of the work here. Verified empirically on this repo:
- It forces LF in the index **and** working tree for all text files (incl. `pnpm-lock.yaml`), overriding each contributor's `core.autocrlf`/`core.eol`.
- It auto-detects binaries by content and skips normalization with **no** explicit rule (a NUL-containing file with only the wildcard resolves to `i/-text w/-text`, identical to an explicit `binary` line).
- `git add --renormalize .` rewrites nothing (the repo is already 100% LF per `.editorconfig` `end_of_line = lf`).

Given that, ~88% of the file is inert for this project:
- **60 `binary` rules match 0 tracked files** (no fonts/archives/audio/video/images/objects/exe/pdf/sqlite exist, and none are plausible in a tiny TS benchmarking lib). `text=auto` already covers the degenerate cases anyway.
- The **lockfile block is redundant** — `text` (Set) vs `text=auto` yields the identical `eol: lf` result, and 3 of the 4 lockfiles don't exist (pnpm-only).
- This is a verbatim copy of a full-stack web `.gitattributes` template (`Web.gitattributes`), which doesn't match this repo's file inventory.

For reference, the comparable minimalist ecosystem (execa, p-limit, vite, vitest, rollup) ships exactly `* text=auto eol=lf` (one line); tinypool/tinyspy ship none.

Requesting changes: please trim to the behaviorally-equivalent minimal version that fits tinybench's minimalist ethos:

```gitattributes
# Normalize line endings across platforms; auto-detect binary by content.
* text=auto eol=lf
```

(Optionally keep `*.patch -text` / `*.diff -text` if you anticipate committing byte-sensitive fixtures.)

.gitattributes
chernojagne: You're right on all counts — none of those extensions are tracked here and `text=auto` handles binary detection by content. Dropped the entire catalogue in be3259c.

.gitattributes
chernojagne: Agreed — removed the whole lockfiles block in be3259c. The wildcard covers `pnpm-lock.yaml` already, and the other entries were dead patterns for this repo.

.gitattributes
chernojagne: Good catch on both — dropped the provenance comment and the `*.map` rule (no tracked map files with `dist/` gitignored) in be3259c.

chernojagne:
Thanks for the detailed review, @jerome-benoit — all points taken.

Updated in be3259c7: stripped the file down to just `* text=auto eol=lf`. Dropped the binary catalogue (matched nothing and `text=auto` handles content-based detection), the redundant lockfile block (wildcard already covers `pnpm-lock.yaml`), and the inert `*.map` rule + provenance comment.

Final file is two lines.

jerome-benoit (approved):

PR #643, opened 2026-08-26
docs(examples): add Bencher result converter

### LINKED ISSUE #182: Documentation: expose benchmark results converters to well known formats
It's suitable for examples code. It will prepare tinybench for usage in CI. 

Formats to consider: 

* BMF: https://bencher.dev/docs/reference/bencher-metric-format/
* Codspeed
* ...

oiahoon:
## Summary

- add a runnable example that emits Bencher Metric Format JSON
- map completed task latency and throughput statistics to the built-in Bencher measures
- include mean and margin-of-error bounds, converting latency from milliseconds to nanoseconds

## Verification

- pnpm typecheck
- pnpm lint
- pnpm lint:typedoc
- pnpm test (241 tests)
- pnpm build
- pnpm --dir examples example:convert-bencher
- parsed the captured output and validated finite values and ordered bounds

Closes #182

### DIFF
--- a/examples/src/convert-bencher.ts
+++ b/examples/src/convert-bencher.ts
@@ -17,6 +17,8 @@ const toMetric = (
   const value = convert(statistics.mean)
   const marginOfError = convert(statistics.moe)
 
+  // The bounds are Tinybench's two-sided 95% Student's t confidence interval
+  // for the sample mean, not sample dispersion or a Bencher threshold.
   return {
     lower_value: Math.max(0, value - marginOfError),
     upper_value: value + marginOfError,
--- a/examples/src/convert-bencher.ts
+++ b/examples/src/convert-bencher.ts
@@ -1,0 +1,57 @@
+import { Bench, mToNs, type Statistics, type Task } from '../../src'
+
+type BencherBenchmark = Record<'latency' | 'throughput', BencherMetric>
+
+interface BencherMetric {
+  lower_value: number
+  upper_value: number
+  value: number
+}
+
+type BencherMetricFormat = Record<string, BencherBenchmark>
+
+const toMetric = (
+  statistics: Statistics,
+  convert: (value: number) => number = value => value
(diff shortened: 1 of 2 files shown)


examples/src/convert-bencher.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type BencherMetricFormat = Record<string, BencherBenchmark>
+
+const toMetric = (
+  statistics: Statistics,
+  convert: (value: number) => number = value => value
+): BencherMetric => {
+  const value = convert(statistics.mean)
+  const marginOfError = convert(statistics.moe)
```
jerome-benoit, line 18: **Suggestion:** Could we document that these bounds represent Tinybench’s two-sided 95% t-confidence interval for the sample mean? BMF adapters use different conventions for `lower_value` and `upper_value`—including standard deviation, IQR, min/max, and confidence intervals—so stating the chosen semantics would prevent this band from being mistaken for sample dispersion or a Bencher regression threshold.

jerome-benoit (dismissed):
The conversion is mathematically consistent: Tinybench’s `moe` is the half-width of a two-sided 95% Student’s t confidence interval for the sample mean. Applying the same linear conversion to the latency mean and margin of error correctly produces nanoseconds, while throughput remains in ops/s.

Bencher does not impose a universal statistical meaning on `lower_value` and `upper_value`, and its regression detection uses `value` rather than these bounds. Therefore, using Tinybench’s MoE here is valid.

No blocking findings.

examples/src/convert-bencher.ts
oiahoon: Addressed in 9dda983. The converter now states that the bounds are Tinybench's two-sided 95% Student's t confidence interval for the sample mean, rather than sample dispersion or a Bencher threshold.

jerome-benoit (approved):

PR #660, opened 2026-09-05
fix: avoid ambient type leaks in declarations

### LINKED ISSUE #658: tinybench@6.1.5 declarations require globalThis.Event and DOMHighResTimeStamp with Wrangler-generated Workerd types
### Description

With `tinybench@6.1.5`, a consumer using runtime declarations generated by `wrangler types` still fails declaration checking because `BenchEvent` extends `globalThis.Event` and `performanceNow` returns `DOMHighResTimeStamp`.

This is a follow-up to #644 and the merged fix #646. The listener-override errors reported in #644 are gone. The two remaining ambient-type errors were already noted in #646; this issue tracks them separately.

### Reproduction

Use the existing public Workerd fixture at this fixed commit and update the three package pins:

https://github.com/screenfluent/upstream-type-repros/tree/2b0ff6d355a8631199c2d6cbe45f397f472ae222/tinybench-workerd

```sh
git clone https://github.com/screenfluent/upstream-type-repros.git
cd upstream-type-repros
git checkout 2b0ff6d355a8631199c2d6cbe45f397f472ae222
cd tinybench-workerd
npm install --save-exact tinybench@6.1.5 wrangler@4.129.0 typescript@6.0.3 --ignore-scripts
npm run reproduce
```

The script runs `wrangler types` followed by `tsc --pretty false`. The consumer is just:

```ts
import { Bench } from "tinybench";
export const bench = new Bench();
```

Its TypeScript configuration uses `lib: ["ES2024"]`, `types: []`, `moduleResolution: "Bundler"`, `strict: true`, and `skipLibCheck: false`, and includes the generated `worker-configuration.d.ts`. It intentionally does not mix browser DOM declarations into the Workerd environment.

### Actual result

```text
node_modules/tinybench/dist/index.d.ts(68,120): error TS2339: Property 'Event' does not exist on type 'typeof globalThis'.
node_modules/tinybench/dist/index.d.ts(1031,37): error TS2304: Cannot find name 'DOMHighResTimeStamp'.
```

The affected published declarations are `BenchEvent ... extends globalThis.Event` and `declare const performanceNow: () => DOMHighResTimeStamp`.

### Expected result

The published declarations should compose with the generated Workerd runtime declarations without requiring `lib.dom` or `skipLibCheck`. If an additional supported configuration is required for this consumer, please clarify it.

### Environment

- Tinybench 6.1.5
- Wrangler 4.129.0; fixture compatibility date 2026-08-25
- TypeScript 6.0.3
- Node 24.19.0, npm 11.19.0

This report concerns declaration compatibility; it does not assert a runtime failure.

michalius:
Fixes #658.

The generated declarations currently leak two ambient types from Tinybench's build environment:

- `BenchEvent` extends `globalThis.Event`
- `performanceNow` is inferred as returning `DOMHighResTimeStamp`

These declarations fail for consumers that intentionally don't include `lib.dom`, including the Workerd reproduction from #658 and strict Node-only consumers.

This change:

- uses `Event` directly for `BenchEvent`
- explicitly types `performanceNow` as the existing `NowFn` (`() => number`)

The runtime behavior is unchanged; the changes only prevent environment-specific ambient types from leaking into the published declarations.

### DIFF
--- a/src/event.ts
+++ b/src/event.ts
@@ -14,7 +14,7 @@ import type {
 class BenchEvent<
   K extends BenchEvents = BenchEvents,
   M extends 'bench' | 'task' = 'bench'
-> extends globalThis.Event {
+> extends Event {
   declare type: K
 
   /**
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -890,7 +890,7 @@ export const withConcurrency = async <R>(
  * Returns the current timestamp in milliseconds using `performance.now()`.
  * @returns the current timestamp in milliseconds
  */
-export const performanceNow = globalThis.performance.now.bind(
+export const performanceNow: NowFn = globalThis.performance.now.bind(
   globalThis.performance
 )
 


michalius:
I noticed #659 was opened shortly before this PR and addresses the same issue more comprehensively, including declaration regression coverage. Closing this one in favor of #659.

Thanks. Glad to have independently confirmed the NowFn fix and the Node/Vitest 5 reproduction in #658.

